PR #1, opened 2019-03-07
add some simple benchmarks as well as a selection from the benchmark game

sadiqj:
Contains a few simple parameterised tests for gc allocation and a selection from the benchmark game


stedolan:
Looks good!

I've been trying to support ocaml back to 4.05 so that we can generate historical performance data. The only thing that breaks on 4.05 with this patch is `List.init` in `stress.ml`, would you mind inlining the definition from the 4.06 stdlib? (https://github.com/ocaml/ocaml/blob/trunk/stdlib/list.ml#L79-L82)

stedolan:
Great, thanks!

PR #2, opened 2019-03-10
More microbenchmarks

sadiqj:
Added simple tests for:

- ocaml-c api
- lazy
- finalisers
- stacks
- weak arrays

### DIFF
--- a/benchmarks/simple-tests/capi.ml
+++ b/benchmarks/simple-tests/capi.ml
@@ -1,10 +1,8 @@
 let test_name = Sys.argv.(1)
 
 let run_many_times f =
-  for _ = 1 to 1_000 do
-    for _ = 1 to 1_000_000 do
+    for _ = 1 to 1_000_000_000 do
       ignore(Sys.opaque_identity f ())
-    done
     done
 
 let () = match test_name with
--- a/benchmarks/simple-tests/capi.ml
+++ b/benchmarks/simple-tests/capi.ml
@@ -1,15 +1,17 @@
 let test_name = Sys.argv.(1)
 
 let run_many_times f =
-    for _ = 1 to 20000 do
-        ignore(f())
+    for _ = 1 to 1_000 do
+        for _ = 1 to 1_000_000 do
+            ignore(f())
+        done
     done
 
 let () = match test_name with
-    | "test_no_args_alloc" -> run_many_times (fun _ -> test_no_args_alloc())
-    | "test_no_args_noalloc" -> run_many_times (fun _ -> test_no_args_noalloc())
-    | "test_few_args_alloc" -> run_many_times (fun _ -> test_few_args_alloc 1)
-    | "test_few_args_noalloc" -> run_many_times (fun _ -> test_few_args_noalloc 2)
-    | "test_many_args_alloc" -> run_many_times (fun _ -> test_many_args_alloc 1 2 3 4 5 6 7)
-    | "test_many_args_noalloc" -> run_many_times (fun _ -> test_many_args_noalloc 1 2 3 4 5 6 7)
+    | "test_no_args_alloc" -> run_many_times (fun _ -> Ocamlcapi.test_no_args_alloc())
+    | "test_no_args_noalloc" -> run_many_times (fun _ -> Ocamlcapi.test_no_args_noalloc())
+    | "test_few_args_alloc" -> run_many_times (fun _ -> Ocamlcapi.test_few_args_alloc 1)
+    | "test_few_args_noalloc" -> run_many_times (fun _ -> Ocamlcapi.test_few_args_noalloc 2)
+    | "test_many_args_alloc" -> run_many_times (fun _ -> Ocamlcapi.test_many_args_alloc 1 2 3 4 5 6 7)
+    | "test_many_args_noalloc" -> run_many_times (fun _ -> Ocamlcapi.test_many_args_noalloc 1 2 3 4 5 6 7)
     | _ -> failwith "unexpected test name"
--- a/benchmarks/simple-tests/capi.ml
+++ b/benchmarks/simple-tests/capi.ml
@@ -3,7 +3,7 @@ let test_name = Sys.argv.(1)
 let run_many_times f =
   for _ = 1 to 1_000 do
     for _ = 1 to 1_000_000 do
-      ignore(f())
+      ignore(Sys.opaque_identity f ())
     done
     done
 
--- a/benchmarks/simple-tests/capi.ml
+++ b/benchmarks/simple-tests/capi.ml
@@ -1,0 +1,15 @@
+let test_name = Sys.argv.(1)
+
+let run_many_times f =
+    for _ = 1 to 20000 do
+        ignore(f())
+    done
+
+let () = match test_name with
+    | "test_no_args_alloc" -> run_many_times (fun _ -> test_no_args_alloc())
+    | "test_no_args_noalloc" -> run_many_times (fun _ -> test_no_args_noalloc())
+    | "test_few_args_alloc" -> run_many_times (fun _ -> test_few_args_alloc 1)
+    | "test_few_args_noalloc" -> run_many_times (fun _ -> test_few_args_noalloc 2)
+    | "test_many_args_alloc" -> run_many_times (fun _ -> test_many_args_alloc 1 2 3 4 5 6 7)
+    | "test_many_args_noalloc" -> run_many_times (fun _ -> test_many_args_noalloc 1 2 3 4 5 6 7)
+    | _ -> failwith "unexpected test name"
--- a/benchmarks/simple-tests/capi.ml
+++ b/benchmarks/simple-tests/capi.ml
@@ -1,10 +1,10 @@
 let test_name = Sys.argv.(1)
 
 let run_many_times f =
-    for _ = 1 to 1_000 do
-        for _ = 1 to 1_000_000 do
-            ignore(f())
-        done
+  for _ = 1 to 1_000 do
+    for _ = 1 to 1_000_000 do
+      ignore(f())
+    done
     done
 
 let () = match test_name with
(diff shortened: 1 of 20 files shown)


benchmarks/simple-tests/finalise.ml
```diff
@@ -0,0 +1,17 @@
+let percent_finalize = int_of_string Sys.argv.(1)
+let words_to_allocate = 1000000000.
+
+type a_record = { an_int: int; mutable a_string : string; a_float : float }
+
+let allocate () =
+  for y = 0 to (if percent_finalize > 0 then percent_finalize*10 else 1000) do
```
stedolan: This logic doesn't make sense to me.
The number of iterations is `percent_finalize*10` and one in every `percent_finalize` of those triggers `Gc.finalise`. For all values of `percent_finalize`, 10% of iterations call `Gc.finalise`.

stedolan:
Looks good! Glad to have these microbenchmarks, it'll be good to have some regression tests for performance.

Two general comments about the implementations:

  - Some of these benchmarks look like a smart optimiser could delete chunks of them. (e.g. `ignore(ref v)` in `finalise.ml`). The standard library has a feature specifically to help with this: `Sys.opaque_identity` is a function that just returns its argument. However, the optimiser does not know that `Sys.opaque_identity` returns its argument, so it has to compute the argument fully. 

    So, instances of `ignore v` should be `ignore (Sys.opaque_identity v)` to guarantee that `v` is computed, and benchmark loops like `for i = 1 to 1000 do ignore (f x) done` should be `for i = 1 to 1000 do ignore (Sys.opaque_identity f x) done`.
G
    There's a subtlety here: it should be `Sys.opaque_identity f x` not `Sys.opaque_identity (f x)` - the latter means that the optimiser can see that `f` is applied to `x`, and can optimise the body of `f` accordingly.

  - Some of the benchmarks adjust their iteration counts depending on Gc stats. (e.g. `while Gc.allocated_words () < x do ... done`). This means that the benchmark will give misleading results on configurations that change the number of allocations: e.g. on 32-bit systems, allocating a `float` takes twice the number of words as on 64-bit systems, and some Flambda configurations can reduce the number of allocations compared to trunk (by e.g. combining allocations or noticing that some are unnecessary). If the benchmark adjusts its iteration count in response, its runtime won't reflect these differences.

benchmarks/simple-tests/finalise.ml
sadiqj: Good find. I'm not 100% sure what was going on there. Fixed.

benchmarks/simple-tests/finalise.ml
```diff
@@ -0,0 +1,17 @@
+let percent_finalize = int_of_string Sys.argv.(1)
+let iterations = 1000000
+
+type a_record = { an_int: int; mutable a_string : string; a_float : float }
+
+let allocate () =
+  for y = 0 to 1000 do
+    let v = { an_int = 5; a_string = "foo"; a_float = 0.0 } in
+    if percent_finalize > 0 && y mod percent_finalize == 0 then
```
stedolan: This is still a bit suspect! if `percent_finalize = 2`, we'll call Gc.finalise `50%` of the time.

benchmarks/simple-tests/morestacks.ml
```diff
@@ -0,0 +1,10 @@
+let bar n =
+  n + 5
+
+let foo () = for a = 0 to 1_000_000 do
+  ignore(Sys.opaque_identity bar a)
+done
+
+let () = for _ = 0 to 10_000 do
+  foo()
+done
```
stedolan: What's this benchmark testing?

benchmarks/simple-tests/stacks.ml
```diff
@@ -0,0 +1,35 @@
...
+    f ()
+  done
+
+let () = let result = match arguments with
+    | "ints-small" -> run_many_times (fun _ -> ints_small 100 depth)
+    | "ints-large" -> run_many_times (fun _ -> ints_large 1 2 3 4 5 6 7 8 9 10 11 12 13 14 15 16 depth)
+    | "floats-small" -> run_many_times (fun _ -> floats_small 100.0 depth)
+    | "floats-large" -> run_many_times (fun _ -> floats_large 1.0 2.0 3.0 4.0 5.0 6.0 7.0 8.0 9.0 10.0 11.0 12.0 13.0 14.0 15.0 16.0 depth)
+    | _ -> failwith "unexpected arguments"
+  in ignore(Sys.opaque_identity result)
```
stedolan: Nice! I think we've been sorely missing this one. Multicore has had some perf bugs with many stack arguments in the past.

benchmarks/simple-tests/weakretain.ml
```diff
@@ -0,0 +1,19 @@
...
+let weak_array_size = int_of_string Sys.argv.(1)
+let weak_percent = int_of_string Sys.argv.(2)
+
+let weak_array = Weak.create weak_array_size
+
+let () = 
+  let v = ref 0 in
+  for i = 1 to 100_000_000 do
+    let t = { an_int = 0; a_string = "foo"; a_float = 0.0 } in
+    if i mod weak_percent == 0 then
```
stedolan: Same issue with percent_finalize above. (I think this number is actually 100/weak_percent)

sadiqj: Yes, they're the wrong way around. I will fix both this and finalise.

benchmarks/simple-tests/morestacks.ml
sadiqj: This was one of the examples in the testing doc, to test the impact of the stack overflow check in the prologue.

benchmarks/simple-tests/finalise.ml
sadiqj: Yes, this has been fixed now.

benchmarks/simple-tests/finalise.ml
```diff
@@ -1,12 +1,12 @@
-let percent_finalize = int_of_string Sys.argv.(1)
+let percent_finalize = 100 / int_of_string Sys.argv.(1)
```
stedolan: I'm afraid the saga of `percent_finalize` continues! The way integer division works, this will run `Gc.finalise` 100% of the time as soon as `percent_finalize > 50` (because `100 / 60 = 1`).

sadiqj: Eurgh. That's awkward. I've replaced this with Random in all the tests that use percentages now.

benchmarks/simple-tests/capi.ml
```diff
@@ -0,0 +1,17 @@
+let test_name = Sys.argv.(1)
+
+let run_many_times f =
+  for _ = 1 to 1_000 do
```
ctk21: Any reason why this is two loops rather than a single one?

benchmarks/simple-tests/morestacks.ml
```diff
@@ -0,0 +1,10 @@
+let bar n =
+  Sys.opaque_identity n + 5
```
ctk21: Do we need both Sys.opaque_identity given that we call bar from within opaque_identity?

benchmarks/simple-tests/stacks.ml
```diff
@@ -0,0 +1,35 @@
...
+  a0 +. (floats_small a0 (n - 1)) *. a0
+
+let rec floats_large a0 a1 a2 a3 a4 a5 a6 a7 a8 a9 a10 a11 a12 a13 a14 a15 n = if n == 0 then
+  1.0 
+else 
+  a0 +. a1 +. a2 +. a3 +. a4 +. a5 +. a6 +. a7 +. a8 +. a9 +. a10 +. a11 +. a12 +. a13 +. a14 +. a15 +. (floats_large a0 a1 a2 a3 a4 a5 a6 a7 a8 a9 a10 a11 a12 a13 a14 a15 (n-1)) *. a0
+
+let run_many_times f =
+  for _ = 1 to 20000 do
+    f () |> ignore
```
ctk21: Does this need Sys.opaque_identity? Also do we want to use just one idiom for wrapping a function in ignore (either f () |> ignore or ignore( f () ) )?

benchmarks/simple-tests/stacks.ml
```diff
@@ -0,0 +1,35 @@
...
+  done
+
+let () = let result = match arguments with
+    | "ints-small" -> run_many_times (fun _ -> ints_small 100 depth)
+    | "ints-large" -> run_many_times (fun _ -> ints_large 1 2 3 4 5 6 7 8 9 10 11 12 13 14 15 16 depth)
+    | "floats-small" -> run_many_times (fun _ -> floats_small 100.0 depth)
+    | "floats-large" -> run_many_times (fun _ -> floats_large 1.0 2.0 3.0 4.0 5.0 6.0 7.0 8.0 9.0 10.0 11.0 12.0 13.0 14.0 15.0 16.0 depth)
+    | _ -> failwith "unexpected arguments"
+  in ignore(Sys.opaque_identity result)
```
ctk21: Do we need this ignore(Sys.opaque_identity result)? Just thinking this test doesn't look the same as test-capi.ml in how it is setup.

benchmarks/simple-tests/capi.ml
sadiqj: Good point, that's a holdover from a previous iteration. Fixed.

benchmarks/simple-tests/stacks.ml
sadiqj: Have made ignore consistent across the tests and added an opaque_identity here.

benchmarks/simple-tests/morestacks.ml
sadiqj: no, you're right. The one in foo should prevent bar getting inlined.

benchmarks/simple-tests/stacks.ml
sadiqj: It's different in test-capi because the c calls might have side effects, so couldn't be optimised away if the result is just thrown away.

That said, there's an opaque identity in the run_many_times so you wouldn't be able to optimise these calls either. I've removed the one highlighted here.

PR #12, opened 2019-04-08
Document how to get multicore and OS X working with Sandmark

ctk21:


stedolan:
I think the underlying issue should be fixed after https://github.com/ocaml-multicore/ocaml-multicore/commit/cb094cbc53c30a801a97f1cb1fb0b0d276d54aaf, so this message is hopefully redundant.

PR #13, opened 2019-04-25
Added lockfree and kcas packages as well as a couple of tests

sadiqj:


benchmarks/multicore-lockfree/test_hash.ml
```diff
@@ -0,0 +1,26 @@
...
+let read_percent = int_of_string Sys.argv.(2)
+
+let () = Random.init 42
+
+let h = Hash.create()
+
+let do_stuff_with_hash () = 
+  for _ = 1 to 100_000 do
+    if Random.int 100 > read_percent then
+      begin
```
ctk21: begin and end here look not needed

PR #14, opened 2019-04-25
Adds two implementations of the Mandelbrot benchmark

sadiqj:
One that forks off workers and the other that uses multicore's domains. This will let us compare the two approaches to parallelism.

It's worth pointing out that mandelbrot is an embarrassingly parallel problem with no shared state and where the bulk of the time is spent iterating so the potential gains here from multicore are minimal.


avsm:
This is a good benchmark though, since at a large number of cores there should still be linear speedup on both backends (assuming sufficient physical core availability)

PR #21, opened 2019-05-09
config to use cloud.drone.io for CI 

ctk21:
Basic CI setup that gets all the dependencies setup and does:
`make ocaml-versions/4.07.1.bench`
fails when this returns a non-zero exit (e.g. a benchmark run failing)


ctk21:
It works, merging!

PR #29, opened 2019-06-07
Let orun handle ocaml programs that use escape sequences in stderr

ctk21:
Let orun handle ocaml programs that use escape sequences in stderr (e.g. compilers)

### DIFF
--- a/orun/dune
+++ b/orun/dune
@@ -6,4 +6,4 @@
 (executable
  (name orun)
  (public_name orun)
- (libraries cmdliner yojson unix wait4))
+ (libraries str cmdliner yojson unix wait4))
--- a/orun/orun.ml
+++ b/orun/orun.ml
@@ -6,6 +6,14 @@ type wait4_result = {
 }
 external wait4 : int -> wait4_result = "ml_wait4"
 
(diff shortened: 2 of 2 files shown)


stedolan:
I don't understand this: how did escape sequences end up here? The GC stats are printed in plain text by the ocaml runtime.

ctk21:
> I don't understand this: how did escape sequences end up here? The GC stats are printed in plain text by the ocaml runtime.

The ocaml runtime isn't inserting them. If the ocaml binary writes the escape sequences to stderr then it throws us. The FStar compiler uses escape sequences to highlight various output on stderr. The first line of GC stats block then looks like:
`\027[;mallocated_words ...`

stedolan:
Ah right, so this patch handles the case where the program writes junk to stderr without a trailing newline, so that the GC output doesn't actually begin at the start of a line? Thanks for the explanation!

ctk21:
> Ah right, so this patch handles the case where the program writes junk to stderr without a trailing newline, so that the GC output doesn't actually begin at the start of a line? Thanks for the explanation!

You're right, we're still open to the program writing something without a newline to stderr and breaking us. However for the use cases I had, the patch above got things working.

PR #30, opened 2019-06-10
Capture command line to JSON log in orun

ctk21:


stedolan:
This will do the wrong thing for commands with spaces or unusual characters in them. Here's a more robust version:

```ocaml
let quotes_needed = function
  | 'a'..'z' | 'A'..'Z' | '0'..'9' | '/' | '_' -> false
  | _ -> true
let quote s =
  match String.iter (fun ch -> if quotes_needed ch then raise Exit) s with
  | () -> s
  | exception Exit -> Filename.quote s
let quote_cmd cmdline = String.concat " " (List.map quote cmdline)
```

ctk21:
> This will do the wrong thing for commands with spaces or unusual characters in them. Here's a more robust version:
> 
> ```ocaml
> let quotes_needed = function
>   | 'a'..'z' | 'A'..'Z' | '0'..'9' | '/' | '_' -> false
>   | _ -> true
> let quote s =
>   match String.iter (fun ch -> if quotes_needed ch then raise Exit) s with
>   | () -> s
>   | exception Exit -> Filename.quote s
> let quote_cmd cmdline = String.concat " " (List.map quote cmdline)
> ```

Good point, I will fix.

PR #33, opened 2019-06-13
Sampling coverage

sadiqj:
This PR adds initial profiling and tracing support to orun.

The intention is to have a separate sandmark iteration on each run that emits profiling, hardware counters and GC instrumentation data for each of the benchmarks to HTML reports or in formats that can easily be analysed. This can enable identifying the cause of a performance regression without having to go through the process of replicating the benchmark environment.

This PR uses the perf_event_open syscall to read profiling events and libdwfl to translate these into file/module/function/line information. It currently relies on LBR in ‘call stack’ mode to capture the call stack though in future it would probably be worth doing DWARF-based unwinding and leverage the branch timing mode of LBR instead.

Interfacing with perf_event_open directly means we can ingest multiple sources of performance data in the same run by having perf emit these to the same ring buffer, e.g having hardware counter sampling and GC instrumentation data from BPF

As it stands this PR emits a hotspot HTML report (example: https://toao.com/tmp/hotspots.html) and a JSON trace suitable for ingestion into the Chrome tracing tool (example: https://toao.com/tmp/trace.json):

image

Additional traces can be added to the output in future to allow slicing up profiling information based on events of interest (GC phases, possibly even high level application events).

To kick off an iteration of orun with profiling enabled, you need to set the environment variable ORUN_CONFIG_PROFILE and instruct sandmark to only do a single iteration (via ITER=1 parameter to the Makefile) as such:

`ORUN_CONFIG_PROFILE=y make ITER=1 ocaml-versions/4.07.1.bench`

You will need to set `/proc/sys/kernel/perf_event_paranoid` to 0 in order to do this as a non-superuser.

Also the default sampling rate is set at 3khz in this PR but we probably want to bump that higher - doing so requires tweaking a few other sysfs settings though.

This also puts a dependency on libdw-dev for sandmark, though this only applies on Linux.

I've talked to @samoht and he's interested in using this with his Irmin benchmarking work, so it may make sense to move some of this code to a separate opam package.

### DIFF
--- a/orun/orun.ml
+++ b/orun/orun.ml
@@ -1,42 +1,54 @@
-type wait4_result = {
-  status: Unix.process_status;
-  user_secs: float;
-  sys_secs: float;
-  maxrss_kB: int
-}
+type wait4_result =
+  { status: Unix.process_status
+  ; user_secs: float
+  ; sys_secs: float
+  ; maxrss_kB: int }
+
 external wait4 : int -> wait4_result = "ml_wait4"
 
 let escape_re = Str.regexp "\\(\027\\[[0-9:;]*m\\)*"
 
-let clean_escape_sequences line =
-  Str.global_replace escape_re "" line
+let clean_escape_sequences line = Str.global_replace escape_re "" line
 
 let escape_sequences_only line =
   let _ = Str.search_forward escape_re line 0 in
   Str.matched_string line
 
 let quotes_needed = function
-  | 'a'..'z' | 'A'..'Z' | '0'..'9' | '/' | '_' | '.' | ',' | '-' | '+' | ':' -> false
-  | _ -> true
+  | 'a' .. 'z'
+  | 'A' .. 'Z'
+  | '0' .. '9'
+  | '/'
+  | '_'
+  | '.'
+  | ','
+  | '-'
+  | '+'
+  | ':' ->
+      false
+  | _ ->
+      true
 
 let quote s =
   match String.iter (fun ch -> if quotes_needed ch then raise Exit) s with
-  | () -> s
-  | exception Exit -> Filename.quote s
+  | () ->
+      s
+  | exception Exit ->
+      Filename.quote s
 
 let quote_cmd cmdline = String.concat " " (List.map quote cmdline)
 
 let starts_with s line =
-  String.length line >= String.length s &&
-    s = String.sub line 0 (String.length s)
+  String.length line >= String.length s
+  && s = String.sub line 0 (String.length s)
 
 let break ch s =
   let open String in
   let i = index s ch in
-  trim (sub s 0 i), trim (sub s (i+1) (String.length s - (i+1)))
+  (trim (sub s 0 i), trim (sub s (i + 1) (String.length s - (i + 1))))
 
 let chop_prefix pfx s =
-  assert (starts_with pfx s);
+  assert (starts_with pfx s) ;
   String.sub s (String.length pfx) (String.length s - String.length pfx)
 
 let get_ocaml_config () =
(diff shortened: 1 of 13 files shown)


kayceesrk:
Are you planning to generate the HTML report as part of the repo? Would be useful to understand what other pieces are left in the PR.

Do you have thoughts on how to get a similar profile for `operf-micro` used in `bench.ocamllabs.io`?

sadiqj:
> Do you have thoughts on how to get a similar profile for `operf-micro` used in `bench.ocamllabs.io`?

This is something I still need to look at. It might actually be easier if we move the profiling code out in to a separate package.

orun/orun.ml
```diff
@@ -82,103 +94,161 @@ let get_ocaml_config () =
...
   (* workaround for the lack of execve *)
-  let prog = if Filename.is_implicit prog && Sys.file_exists prog then
-               "./" ^ prog else prog in
+  let prog =
+    if Filename.is_implicit prog && Sys.file_exists prog then "./" ^ prog
+    else prog
+  in
   try
+    let profiling =
+      match Sys.getenv_opt "ORUN_CONFIG_PROFILE" with
```
ctk21: Not a fan of using an environment variable to set the behaviour of the program. However I see the problem we have with the dune files launching orun. 

We could land this and sort out a way to configure the dune rules to pass config switches later.

orun/orun.ml
```diff
@@ -82,103 +94,161 @@ let get_ocaml_config () =
...
+        | "mean_space_overhead" ->
+            let value = float_of_string value in
+            (key, `Float value) :: go true
+        | _ ->
+            let value = int_of_string value in
+            (key, `Int value) :: go true )
+  in
   `Assoc (go false)
 
+let exec_prog profiling output_name prog cmdline env stdin stdout stderr =
```
ctk21: This feels wrong. 
Shouldn't we have two functions:
 - one that calls `Profiler.create_process_env_paused ... result; pid)`
 - and the raw `Unix.create_process_env ... `
Then have the config switch setup the evaluation function to be one or the other?

orun/orun.ml
```diff
@@ -1,42 +1,54 @@
...
+let clean_escape_sequences line = Str.global_replace escape_re "" line
 
 let escape_sequences_only line =
   let _ = Str.search_forward escape_re line 0 in
   Str.matched_string line
 
 let quotes_needed = function
-  | 'a'..'z' | 'A'..'Z' | '0'..'9' | '/' | '_' | '.' | ',' | '-' | '+' | ':' -> false
-  | _ -> true
+  | 'a' .. 'z'
```
kayceesrk: Was this reformatted by ocamlformat?

orun/orun.ml
```diff
@@ -1,42 +1,54 @@
...
 
 let break ch s =
   let open String in
   let i = index s ch in
-  trim (sub s 0 i), trim (sub s (i+1) (String.length s - (i+1)))
+  (trim (sub s 0 i), trim (sub s (i + 1) (String.length s - (i + 1))))
 
 let chop_prefix pfx s =
-  assert (starts_with pfx s);
+  assert (starts_with pfx s) ;
```
kayceesrk: whitespace

orun/orun.ml
sadiqj: Yea, I think I put it through ocamlformat before committing.

orun/orun.ml
```diff
@@ -82,103 +94,161 @@ let get_ocaml_config () =
...
+      else output
+    in
+    let environ =
+      "OCAMLRUNPARAM=v=0x400"
+      :: List.filter
+           (fun s -> not (starts_with s "OCAMLRUNPARAM="))
+           (Array.to_list (Unix.environment ()))
+    in
+    let pid =
+      exec_prog profiling name prog (Array.of_list cmdline)
```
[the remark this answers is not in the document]
sadiqj: Good idea, this should be fixed in a sec.

orun/orun.ml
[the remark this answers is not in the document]
sadiqj: I agree, environment variable isn't a good way to do this but it seemed least-worst for now.

A solution to configure config switches is where it should fit in eventually.

orun/orun.ml
sadiqj: Have pushed a version that definitely went through ocamlformat.

PR #36, opened 2019-07-09
add tests for hashtbl, array, stack, set, map, str, string, bytes and pervasives.

sadiqj:


benchmarks/data-structures/hashtbl_bench.ml
```diff
@@ -0,0 +1,76 @@
...
+
+  let hash i = i land max_int
+end
+
+module IntHashtbl = Hashtbl.Make (IntHash)
+
+let gen_test_int_replace_and_find n tbl =
+  let replace () =
+    for i = 0 to n - 1 do
+      Hashtbl.add tbl i i
```
ctk21: do we mean `Hashtbl.replace` here? (`add` and `replace` are actually different)

benchmarks/data-structures/hashtbl_bench.ml
```diff
@@ -0,0 +1,76 @@
...
+      let (_ : int option) = Hashtbl.find_opt tbl i in
+      ()
+    done
+  in
+  (replace, find)
+
+let gen_test_int_replace_and_find_int_hash n tbl =
+  let replace () =
+    for i = 0 to n - 1 do
+      IntHashtbl.add tbl i i
```
ctk21: do we mean `Hashtbl.replace` here? (`add` and `replace` are actually different)

benchmarks/data-structures/hashtbl_bench.ml
sadiqj: Well spotted. Fixed.

PR #39, opened 2019-07-18
split out multibench target in to serial and parallel

sadiqj:


ctk21:
something bad with the CI here, but we want to drop this to progress #54 
we are tracking it as an issue #58

PR #50, opened 2019-09-11
dune.inc approach to run perf tools

snimmagadda:
As discussed, this PR implements a new 'bench-run' action via dynamically generating dune.inc file within each benchmark sub-directory.

For this iteration, let's consider 'BENCH_TARGET=benchmarks/bdd/bench-run' alone as some benchmarks have inconsistent executable names, for example benchmarks/sequence generates 'sequence-cps.exe' instead of 'sequence.exe' which prevents batch generation of dune.inc files.

make BENCH_TARGET=benchmarks/bdd/bench-run ocaml-versions/4.07.1.bench

will now generate a new rule within the empty dune.inc files with a hard coded 'perf stat -ax'|'' command for now and upon completion of execution will generate bdd.perf files within the respective _build subdirectories.

 The perf files could be post processed as needed or concatenated together  just like .bench files.

Concerns about this approach...
1. dune requires dune.inc files to be present on the filesystem and cannot be autogenerated.
2. Which requires empty dune.inc to be checked in to the tree. (otherwise dune errors out for (include dune.inc) stanza).
3. Upon execution, these auto-generated rules in dune.inc will show up on 'git diff' which will be a noise as these needs to be overwritten for every run and shouldn't be checked-in.
4. Couldn't ignore them in .gitignore as they are tracked files.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -84,6 +84,7 @@ ocaml-versions/%.bench: ocaml-versions/%.comp _opam/% .FORCE
 	   for i in `seq 1 $(ITER)`; do \
 	     echo "(context (opam (switch $*) (name $*_$$i)))"; \
            done } > ocaml-versions/.workspace.$*
+	sh generate_dune_inc.sh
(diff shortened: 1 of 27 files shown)


ctk21:
dropping in favour of #54

PR #51, opened 2019-09-11
Run perf tool in addition to default build action.

snimmagadda:
This is another approach of generating perf stats in addition to the default
bench stats by running 'perf' right after 'orun' during bench build rule.

Pros:
Simple and mechanical diff.
Any additions tools/generators could be concatenated to the list of actions).

Cons:
Hard coded executable and command line parameters.

### DIFF
--- a/benchmarks/bdd/dune
+++ b/benchmarks/bdd/dune
@@ -6,6 +6,9 @@
 (rule
  (targets bdd.bench)
  (deps (:prog bdd.exe))
- (action (run orun -o %{targets} -- %{prog})))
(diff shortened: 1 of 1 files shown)


ctk21:
Sorry, I don't think this approach is going to work well in practice. When running backfills and tests the runtime is already large. This patch hardcodes it to be longer without an easy way to configure the command line params. 

We need an easy way to allow the user to configure what is run without hardcoding.

ctk21:
Dropping this to focus on approach in #54

PR #53, opened 2019-09-12
adds base

sadiqj:


kayceesrk:
@sadiqj do we have benchmarks in base library that we can add here.

sadiqj:
> @sadiqj do we have benchmarks in base library that we can add here.

No, we parked this PR because I think we were worried about trunk breaking continuously. It probably wouldn't be hard to port the tests from `stdlib` over.

kayceesrk:
Ok. Better to leave it as it is for now.

shakthimaan:
The following PR https://github.com/ocaml-bench/sandmark/pull/132 updates base to v0.14.0 so that Sandmark can build with dune-2.5.1 and Multicore OCaml 4.10.0, and hence this PR can be closed.

PR #55, opened 2019-09-18
Proof of concept: Bring orun out of dune rules.

snimmagadda:
For this diff let's consider BENCH_TARGET=benchmarks/bdd alone.

The master branch with 'make ocaml-versions/4.07.1.bench
BENCH_TARGET=benchmarks/bdd/bench' and this PR with 'make
ocaml-versions/4.07.1.bench BENCH_TARGET=benchmarks/bdd' (note just
directory without a bench target) produces similar[0] output in
ocaml-versions/4.07.1.bench.

[0] Full path of executable with this PR vs just basename with master.


ctk21:
dropping this to focus on approach in #54

PR #59, opened 2019-10-07
Add a note about running sandmark from non-standard locations

Engil:
This may add a bit of noise in README.md, if you think it's ok then feel free to merge it, otherwise I can move it or shorten it. :)


ctk21:
Looks good to me. Thanks for chasing it through!

PR #63, opened 2019-10-14
Add required parameter for orun

gs0510:
The man page for orun says(doesn't mention -o FILE is required):
`        -o FILE, --output=FILE
           Output location for run statistics
`

and the orun errors out with the following error message:

`
orun: internal error, uncaught exception:
      Sys_error(": No such file or directory")
`

Adding required parameter to -o so that it fails with a more meaningful message if output file is not present.

### DIFF
--- a/orun/orun.ml
+++ b/orun/orun.ml
@@ -259,7 +259,7 @@ open Cmdliner
 
 let output =
   let doc = "Output location for run statistics" in
-  Arg.(value & opt string "" & info ["o"; "output"] ~docv:"FILE" ~doc)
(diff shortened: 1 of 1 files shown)


orun/orun.ml
```diff
@@ -259,7 +259,7 @@ open Cmdliner
 
 let output =
   let doc = "Output location for run statistics" in
-  Arg.(value & opt string "" & info ["o"; "output"] ~docv:"FILE" ~doc)
+  Arg.(required & value & opt string "" & info ["o"; "output"] ~docv:"FILE" ~doc)
```
ctk21: I don't think arguments like this can be tagged 'required'; the CI fail seems to suggest this.

PR #68, opened 2019-11-21
Add the ability to run multicore parallel benchmarks.

kayceesrk:
+ Split the multicore benchmarks into multicore parallel (those which
use Domains) and multicore serial (those which only use effect
handlers).
+ Introduce `multicore_parallel_run_config.json` -- a new run config
json file for multicore parallel benchmarks
+ Change the default number of iterations in the Makefile to 1 as this
is what is used by ocaml-speed
+ Cosmetic changes to README.md


ctk21:
LGTM

PR #83, opened 2020-02-24
Graph algorithm

atuldhiman:
Add LU decomposition.


sadiqj:
I'm a little bit confused by the `serial_versions` directory contents here, most of these benchmarks are already in `multicore-numerical`? Also there doesn't seem to be a dune file or anything in the `serial_versions` directory so none of them build.

If the intention is to separate out the serial and multicore benchmarks - could we put all the serial ones in `serial-numerical` (including those currently in `multicore-numerical`)?

sadiqj:
`LU_decomposition_multicore` and `LU_decomposition_serial` are missing from the `dune` file and so neither of those build. Would it be possible to drop the `_serial` suffix from these so they are consistent with the other benchmarks (which have _multicore and no suffix for serial)?

sadiqj:
`floyd_warshall` and `floyd_warshall_multicore` are missing from the `dune` file `multibench_parallel` alias. Poke me on slack, I have an updated `dune` where they build.

kayceesrk:
@Sudha247 can you please work with @atuldhiman to fix the build based on Sadiq's comments.

PR #84, opened 2020-02-25
quicksort median of three parallel

Sudha247:


kayceesrk:
As discussed offline, there's a dirtier way of finding medians which will work for quicksort. @Sudha247 rather than a new benchmark, modify the existing one to use the new median finding algorithm.

kayceesrk:
We're probably not going to merge this one. There are better sorting algorithms that we should explore.

PR #93, opened 2020-03-05
Adding irmin

anmolsahoo25:
This PR adds the dependencies required for irmin to sandmark. This should enable us, going forward, to write irmin benchmarks.


run_config.json
```diff
@@ -1106,6 +1106,16 @@
           "params": ""
         }
       ]
+    },
+    {
+      "executable": "benchmarks/irmin/basic_read_write.exe",
+      "name": "basic_read_write",
+      "ismacrobench": true,
```
kayceesrk: The benchmark doesn't do much work. Hence, this should be `false`.

kayceesrk:
Looks like the CI failure is due to an earlier commit.

anmolsahoo25:
The build fails because I have created 2 versions of `ctypes` - 
1. `cyptes.0.14.0`
2. `ctypes.0.14.0+multicore`

The `--best-effort` switch always forces `ctypes.0.14.0+multicore` option even if `ctypes.0.14.0` exists. I propose removing the `--best-effort` switch and checking if everything works, which should fix this patch.

PR #98, opened 2020-03-09
Add the read and write irmin benchmark

shubhamkumar13:
Add the basic_read_write file in the irmin benchmark


run_config.json
```diff
@@ -1113,7 +1113,10 @@
       "ismacrobench": false,
```
kayceesrk: Can you rename this to `irmin_mem_rw` instead of `basic_read_write`

shubhamkumar13: Yes sure, that seems more fitting.

kayceesrk: The commit days comment out necessary print statements. But I don’t see any print statements commented out. Can you ensure that those files are not missed out.

shubhamkumar13: I checked the files and they are not missed out.
I made a mistake in the commit message, those print statements were present on the https://github.com/ocaml-bench/sandmark/pull/98/commits/56e64294ab9d33f067b9c12ce52c8ebe5f51eb95 commit. At that time I was probably thinking of squashing them
I'm renaming the most recent commit message.

PR #100, opened 2020-03-20
Gram matrix initialization numerical benchmark

Sudha247:


kayceesrk:
Looks good!

UnixJunkie:
@Sudha247 you can put a comment in the header of the files, stating that you are the author of the multicore version. I feel bad if there is only my name at the top of all those files.
Thanks a lot for the port, by the way.

PR #102, opened 2020-03-23
add gramMatrix to bench

Sudha247:
Adds gramMatrix benchmark to `multibench_parallel` and `build_bench`, fixes input file location.


kayceesrk:
The correct way to handle path is to add the input files as dependencies in the dune file. I've merged the correct change here: https://github.com/ocaml-bench/sandmark/commit/1ce3e5038c931fa3b85b7badbaed3295fe4c3a97

Issue #101, opened 2020-03-23
sandmark requires the system package libdw-dev on Ubuntu


UnixJunkie:
Since there is no opam file, its installation is not automatic.

Maybe, there should be an opam file.

kayceesrk:
The easiest way now would be to add a `Make` rule called `depend` which installs `libdw-dev`. I'm not sure if an OPAM file is the right way to install system dependencies. @shakthimaan would you be able to do this?

shakthimaan:
Sure, but, the package names are specific to .deb based distributions. Actually, you require both `libgmp-dev` and `libdw-dev` to be installed on Ubuntu, and this is mentioned in the "pre-requisities" section in the README file.

UnixJunkie:
opam is the right way to install system dependencies.
E.g. in an opam file:
```
depexts: [
  ["gnuplot-x11"] {os-family = "debian"}
  ["gnuplot"] {os-distribution = "centos"}
  ["gnuplot"] {os-distribution = "fedora"}
  ["gnuplot"] {os = "macos" & os-distribution = "homebrew"}
  ["gnuplot"] {os-distribution = "alpine"}
  ["gnuplot"] {os-family = "suse"}
]
```
Allows to create package installs depending on the Linux distro.
Supporting Debian and Ubuntu would be the very minimum.
Usually, in opam, conf-* packages are full of such statements.

shakthimaan:
I shall add an opam file for sandmark shortly. If you already have one that we can use, please feel free to create a PR. Thanks!

UnixJunkie:
I don't have one, sorry.

kayceesrk:
We also need to ensure python3 and pip3 is installed. And also intervaltree library. 

```bash
$ sudo apt install python3 python3-pip
$ pip3 install intervaltree
```

kayceesrk:
Addressed by #103 and #104.

PR #103, opened 2020-03-26
Added depend target in Makefile

shakthimaan:
Issue: https://github.com/ocaml-bench/sandmark/issues/101

Initially, I thought of using the dune, and dune-project to generate the .opam file, but, the Makefile is actually current and maintained. Hence, I have updated the same to include dependency checks for both libgmp-dev and libdw-dev for Ubuntu.


kayceesrk:
Ok. let's go with this for now.

PR #104, opened 2020-03-30
Added python pip3 dependency

shakthimaan:
Updated Makefile to include python and pip3 dependencies as mentioned in https://github.com/ocaml-bench/sandmark/issues/101.


kayceesrk:
Thanks.

PR #109, opened 2020-04-27
Added sequential-interactive.ipynb

shakthimaan:
A first version of the sequential-interactive.ipynb has been created:
* Prompt for the number of bench files for comparison.
* UI to select the host, repository, commit, variant.
* If you are running locally, update the `artifacts_dir` path accordingly.
* The output is not stored in the file to keep the file size to a minimum.

The section from `Selection example` will need to be addressed, as we cannot use explicitly `4.06.1+stock` and the functions that use the same and glob.glob.

Note: You will need to have done `pip install nested_dict` to use the notebook.


shakthimaan:
* Moved notebook to interactive/ folder.
* Added test/artifacts/ folder that can be used to test with papermill.
* Set default comparisons to two instead of zero.
* Add check to print if the number of bench files is less than zero.

kayceesrk:
Are the image files necessary to be in the `sequential` directory? https://github.com/ocaml-bench/sandmark/blob/0b4a54f76031e0078616921ea28da8f363592afd/notebooks/sequential/images/sequential-select-baseline-variant.png

shakthimaan:
> Are the image files necessary to be in the `sequential` directory?
I have now moved them to the interactive/ folder as the README uses the same. Thanks!

kayceesrk:
Merged. Looking forward to having this run on `winter`!

PR #114, opened 2020-05-08
Add 4.10.0+multicore

Sudha247:
This PR adds OCaml versions `4.10.0+multicore` along with `4.10.0+stock` and `4.10.0+stock+instrumented`. 

Also, some minor tweaks in the gram-matrix benchmark to make generating graphs hassle free.


kayceesrk:
Great. Thanks. That grammatrix thing was on my list for a while.

PR #116, opened 2020-05-13
Use sudo for parallel benchmark builds

shakthimaan:
Use sudo with specific environment variables to make opam and dune happy to build parallel compiler and benchmarks. Both `run_all_parallel.sh` and `run_all_serial.sh` build fine.


kayceesrk:
Great. Thanks! Should we enable parallel benchmarks to be built in the CI?

shakthimaan:
> Great. Thanks! Should we enable parallel benchmarks to be built in the CI?

Sure. Will need to ensure that there is enough storage space though.

PR #118, opened 2020-05-18
Refactored README and added JupyterHub info

shakthimaan:
Updated README with JupyterHub information and reorganized the content a bit.


kayceesrk:
Thanks!

PR #121, opened 2020-05-19
Mention sudo access for run_all_parallel.sh script

shakthimaan:
$subject

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,8 +90,9 @@ You can execute both serial and parallel benchmarks using the
 Ensure that the respective .json configuration files have the
 appropriate settings.
 
-The run_all_parallel.sh script uses chrt and hence sudo permission is
-required for the user executing the script.
+The run_all_parallel.sh script uses chrt and the user executing the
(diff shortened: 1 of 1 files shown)


kayceesrk:
Does run all parallel just needs sudo or sudo with nopasswd? The difference is important since it determines whether the process needs an interactive session. When using without nopasswd, when is the password requested? At the start or after the build phase has completed but before the run phase?

shakthimaan:
Since we want the script to be run non-interactively (such as for nightly builds), we would like the user running the script to have "sudo with nopasswd". With this option the password is not requested at all. Moreover, even for multiple compiler builds, it might be a while before a parallel compiler and its benchmark builds complete, and we do not want to interrupt the script execution and wait for the user to enter the sudo password. The use of sudo is minimized though to only one command in the Makefile when it is required to use chrt.

kayceesrk:
Ok. I think that it would be easier to make sudo access with `nopasswd` a requirement for running parallel benchmarks.

kayceesrk:
Thanks!

PR #124, opened 2020-05-28
user configurable paramwrapper added to Makefile

shubhamkumar13:


README.md
```diff
@@ -174,3 +174,20 @@ The ocaml-update-c command in multicore needs to run with GNU
 sed. `sed` will default to a BSD sed on OS X. One way to make things
 work on OS X is to install GNU sed with homebrew and then update the
 `PATH` you run sandmark with to pick up the GNU version.
+
+## User configurable benchmark wrapper
+
+To run a macro benchmark with a custom range of processors a conditional string can be passed with `PARAMWRAPPER`.
```
shakthimaan, line 180: s/custom range of processors/isolated CPUs/

Makefile
```diff
@@ -1,4 +1,14 @@
 # options for running the benchmarks
+#
+# paramwrapper for passing taskset and chrt details
+# When passing arguments to PARAMWRAPPER do follow the syntax 
+# if params < {number of cores} then paramwrapper = {range of cores} else paramwrapper = {range of cores}
```
shakthimaan, line 5: Why not simply specify a range of cores to use instead of a `if` condition?

README.md
```diff
@@ -174,3 +174,20 @@ The ocaml-update-c command in multicore needs to run with GNU
...
+
+For example :
+```bash
+$ make run_config_macro.json PARAMWRAPPER="if params < 16 then paramwrapper = 2-15 else paramwrapper = 2-15,16-21"
+```
+In the above example strings : `16`, `2-15`, `2-15,16-21` are used to construct a json file containing a `paramwrapper` record with the value : `taskset --cpu-list 2-15 chrt -r 1` or `taskset --cpu-list 2-15,16-21 chrt -r 1`. This depends upon `params`, in the above example it happens on `16`. 
+
```
shakthimaan, line 187: Can you please re-phrase the last sentence?

README.md
```diff
@@ -174,3 +174,20 @@ The ocaml-update-c command in multicore needs to run with GNU
...
+The command above generates a new `.json` file. In this example it's `run_config_macro.json`.
```
shakthimaan, line 188: s/it's/it is/

README.md
```diff
@@ -174,3 +174,20 @@ The ocaml-update-c command in multicore needs to run with GNU
...
+
+There are limitations to it's flexibility and it currently only supports that single string. If no optional string is provided it defaults to
```
shakthimaan, line 190: You can probably remove the sentence, `There are limitations ...`.

README.md
shubhamkumar13: Done. I think this looks easier to understand.

Makefile
shubhamkumar13: The `if` condition is more intuitive, where instead of passing 3 arguments we can only pass one.

README.md
```diff
@@ -174,3 +174,20 @@ The ocaml-update-c command in multicore needs to run with GNU
...
 work on OS X is to install GNU sed with homebrew and then update the
 `PATH` you run sandmark with to pick up the GNU version.
+
+## User configurable benchmark wrapper
+
+To run a macro benchmark with a custom range of processors/isolated CPUs/ a conditional string can be passed with `PARAMWRAPPER`.
+
+For example :
+```bash
+$ make run_config_macro.json PARAMWRAPPER="if params < 16 then paramwrapper = 2-15 else paramwrapper = 2-15,16-21"
```
Sudha247, line 184: `run_config.json` contains the sequential benchmarks, which are run on a single core. `multicore_parallel_run_config.json` might be a better example here.

shubhamkumar13: That's a nice suggestion, I've modified the filename. Thanks!

Sudha247:
I had an offline discussion with @shubhamkumar13, it appears to be the case that the syntax works on `bash` shell and my machine is picking up `sh` by default, hence the error.

avsm:
@Sudha247 what distro are you running on? If bash is needed, the makefile should run it under `bash <script>` specifically.

Sudha247:
@avsm I'm running on Ubuntu 18.04.

shubhamkumar13:
I checked with @Sudha247 offline and it seems to work fine on her machine. @shakthimaan can this PR be closed?

Makefile
shakthimaan: Why not just pass only "{range of cores}"  as an argument, say "2-13", or "2-13,16-27"? For a given machine, you already know which cores to use, and you are specifying them to the wrapper script. The use of the `if` condition becomes redundant in this case.

shakthimaan:
Why not just pass only "{range of cores}" as an argument, say "2-13", or "2-13,16-27"? For a given machine, you already know which cores to use, and you are specifying them to the wrapper script. The use of the if condition becomes redundant in this case. The README looks fine to me.

shubhamkumar13:
@shakthimaan I think for now the `if` statement is ok since it's not going to be used extensively.
And will be updated if it has to cover something which is more generic.

shubhamkumar13:
@shakthimaan this PR can be merged now, right? If there is nothing more to add.

PR #125, opened 2020-06-01
Add cubicle benchmarks

Sudha247:
This PR adds two sequential benchmarks of [`cubicle`](http://cubicle.lri.fr/) model checker. They are 

* `german_pfs` - typically runs for about 5 minutes
* `szymanski_at` - typically runs for about 10 minutes

The results for these two benchmarks are [here](https://gist.github.com/Sudha247/cb977c9754955ccc6064482b83029f87).


kayceesrk:
This looks great. Thanks! 

Thanks for adding the notebook. It was quite useful to see the results.

Issue #127, opened 2020-06-08
CI should run parallel benchmarks


kayceesrk:
Currently we only run the sequential benchmarks in the CI. We should run parallel benchmarks as well. It is useful just to run the benchmark for 2 domains, removing the taskset and chrt commands. This can be done with `jq`.

PR #129, opened 2020-06-11
Add target to run parallel benchmarks in the CI

shubhamkumar13:
https://github.com/ocaml-bench/sandmark/issues/127

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -9,12 +9,32 @@ steps:
 - name: 4.07.1.bench
   image: ocaml/opam2:4.07
   commands:
-  - sudo apt-get update && sudo apt-get -y install libgmp-dev m4 libdw-dev
+  - sudo apt-get update && sudo apt-get -y install libgmp-dev m4 libdw-dev && sudo apt-get install jq
+  - sudo chown -R opam .
+  - eval $(opam env)
+  - opam install dune.1.11.4
+  - export ITER=1
+  - export OPAM_DISABLE_SANDBOXING=true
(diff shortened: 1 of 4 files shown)


kayceesrk:
In the same PR, can you also modify the CI scripts to run the parallel benchmarks since that's the end goal? You may modify the PR title appropriately.

.drone.yml
```diff
@@ -15,4 +15,6 @@ steps:
   - opam install dune.1.11.4
   - export ITER=1
   - export OPAM_DISABLE_SANDBOXING=true
-  - make ocaml-versions/4.07.1.bench
```
kayceesrk, line 18: I think we want to run the serial benchmarks as well.

kayceesrk:
We should probably run the parallel benchmark as a separate build. There's no need to run them in the same build as the sequential benchmarks.

.drone.yml
```diff
@@ -15,4 +15,6 @@ steps:
...
+  - make multicore_parallel_run_config_macro_parallel.json
+  - RUN_CONFIG_JSON=multicore_parallel_run_config_macro_parallel.json make ocaml-versions/4.10.0+multicore+parallel.bench
```
Sudha247, line 19: Since the default is `buildbench`  we'll have to explicitly specify `BUILD_BENCH_TARGET=multibench_parallel`  to run the parallel benchmarks.

Sudha247:
It might be worth adding the generated json file to `.gitignore`.

PR #131, opened 2020-06-22
Update decompress benchmarks

dinosaure:
A fix about mirage/decompress#86. I added some notes where, about compression, we can take the advantage of multicore - however I let you to see details about that.

Issue #17, opened 2019-05-02
frama-c benchmarks fail on multicore (build and run)


ctk21:
frama-c benchmarks fail to build on multicore:
 - ctypes is a dependency (but see https://github.com/ocamllabs/sandmark#ctypes for workaround)
 - it uses the name `effect` which is a keyword in multicore. Patch to get going in sandmark here: https://github.com/ocamllabs/sandmark/tree/bugfix/ctk21/frama-c-multicore-effects 

Even with frama-c building, the binary segfaults on multicore. There is a multicore for this: https://github.com/ocaml-multicore/ocaml-multicore/issues/246

PR #132, opened 2020-06-24
Update dependency packages to use dune.2.6.0 and Multicore OCaml 4.10.0

shakthimaan:
The attached patch updates Sandmark to use dune 2.6.0 for Multicore OCaml 4.10.0.

* The dependencies/packages/ have been cleaned up so that only the latest version is retained.
* Additional dependencies for newer releases have been added.
* Both serial and parallel benchmarks run fine (except for frama-c, which is already a known issue. See https://github.com/ocaml-bench/sandmark/issues/17)

In order to test, you need to first update dune locally on your system using:
```
$ opam install dune.2.6.0
```
You will need to apply this patch and the PR from https://github.com/ocaml-bench/sandmark/pull/131 as well in order to build and test. You can obtain the raw patch from a PR by suffixing with .patch. 
For example: https://patch-diff.githubusercontent.com/raw/ocaml-bench/sandmark/pull/131.patch. 

You can then proceed to build the dependencies and run the serial benchmarks using the following command:
```
$ make ocaml-version/4.10.0+multicore.bench
```

In order to run the parallel benchmarks, you can use the following in run_all_parallel.sh, and execute the script:
```
make multicore_parallel_run_config_macro.json

BUILD_BENCH_TARGET=multibench_parallel \
	RUN_CONFIG_JSON=multicore_parallel_run_config_macro.json \
	make ocaml-versions/4.10.0+multicore.bench
```


Makefile
```diff
@@ -27,9 +27,10 @@ CONTINUE_ON_OPAM_INSTALL_ERROR ?= true
 WRAPPER = $(subst run_,,$(RUN_BENCH_TARGET))
 
 PACKAGES = \
-	cpdf menhir minilight camlimages yojson lwt alt-ergo zarith \
-	js_of_ocaml-compiler uuidm react ocplib-endian nbcodec checkseum decompress \
-	sexplib0 irmin-mem cubicle
+       cpdf conf-pkg-config conf-zlib bigstringaf decompress camlzip menhirLib menhir \
+       minilight base stdio dune-private-libs dune-configurator camlimages yojson lwt \
+       alt-ergo zarith integers js_of_ocaml-compiler uuidm react ocplib-endian        \
+       nbcodec checkseum sexplib0 irmin-mem
```
Sudha247, line 33: If you add cubicle to this list of packages, I reckon the cubicle serial benchmarks should run.

kayceesrk:
`dune.2.6.0` should work now thanks to https://github.com/ocaml-multicore/multicore-opam/pull/22

Makefile
shakthimaan: I have added cubicle to PACKAGES in the Makefile and the serial benchmarks ran fine. Thanks!

shakthimaan:
The following dependency packages have been upgraded to build with dune.2.6.0:
* dune.2.6.0
* dune-private-libs.2.6.0
* dune-configurator.2.6.0
* digestif.0.8.1
* checkseum.0.2.1

kayceesrk:
The build is still failing. Should we wait until it turns green for reviewing?

shakthimaan:
The CI is using ocaml/opam2 Docker container from DockerHub:
1. The ocaml/opam2:{latest, 4.12 4.10} images do not have the required dune.2.6.0 version.
2. The packages are pulled from git://github.com/ocaml/ocaml-beta-repository, while dune.2.6.0 version is available only from https://opam.ocaml.org.
Hence, the CI build is failing.

shakthimaan:
I have updated .drone.yml to use ocurrent/opam:debian-10-ocaml-4.10 Docker image that can install dune.2.6.0. Since this PR also requires #131, I have temporarily applied that in the YAML file for the CI to pass. 
```
- wget https://github.com/ocaml-bench/sandmark/pull/131.patch
- patch -p1 < 131.patch
```
When both #131 and this PR are merged, I shall create a PR to remove the patch application from the .drone.yml. file.

.drone.yml
```diff
@@ -6,16 +6,18 @@ platform:
...
   - eval $(opam env)
-  - opam install dune.1.11.4
+  - opam update
+  - opam install dune.2.6.0
   - export ITER=1
   - export OPAM_DISABLE_SANDBOXING=true
-  - make ocaml-versions/4.07.1.bench
+  - wget https://github.com/ocaml-bench/sandmark/pull/131.patch
+  - patch -p1 < 131.patch
+  - make ocaml-versions/4.10.0+multicore.bench
```
ctk21, line 21: Are we able to keep a CI build against stock OCaml 4.10 in here as well?

shakthimaan: We can create multiple `exec pipelines` in the same .drone.yml. Any specific reason to have it as part of this code base, rather than in stock OCaml 4.10 or our clone of the repository? The status of the build is determined by the success of both the pipeline runs though.
Source: See: https://docs.drone.io/pipeline/configuration/#multiple-pipelines

ctk21: I'm happy for it to be in the same run or in another run (I think in the past we had two runs, one for stock and one for multicore). I'm just keen to have CI coverage in case we accidentally make a change to packages which works with multicore but not with stock.

kayceesrk:
As discussed, we need 3 pipelines,

* 4.10.0 & serial benchmarks
* 4.10.0+multicore & serial benchmarks
* 4.10.0+multicore & parallel benchmarks

shakthimaan:
The 4.10.0+stock build got killed because it reached the 60 minute default timeout setting. I tested the build on another machine and the benchmarks ran fine though. A GitHub Administrator will have access to the Drone configuration settings, and the default repo timeout should be increased to at least two hours.
Reference: https://docs.drone.io/server/user/admin/#capabilities

kayceesrk:
@ctk21 Do you have admin access to Drone?

ctk21:
> @ctk21 Do you have admin access to Drone?

We don't have admin access. It is provided as a cloud service by Drone. I think we need to find a way for the run to be under 60mins (which is likely going to be a headache anyhow). 

Do we know what is taking the time? Maybe we need a cut-down config that can give us coverage of the places most likely to go wrong; this feels like the build stages of the benchmarks rather than the execution of all the benchmarks.

shakthimaan:
I have now added a DRY_RUN parameter to the Makefile:
* When its value is set to 1, it runs everything except running the benchmarks. 
* By default, the benchmarks are executed.

kayceesrk:
@shakthimaan could you summarize what is pending to get this PR merged?

shakthimaan:
The PR can be merged to master:
* We have tested the same for building Coq, and it works fine for dune.2.6.0 and Multicore OCaml 4.10.0. 
* I have now updated the branch to reflect the README and Makefile changes from master, and the CI build is clean.
This PR does require the following PR `Update decompress benchmarks` to be merged first as a pre-requisite.
https://github.com/ocaml-bench/sandmark/pull/131

PR #139, opened 2020-07-14
Modified paramwrapper to ensure it to run with serial benchmarks

shubhamkumar13:


kayceesrk:
Can we do it properly with `num_domains` in the `*.json` files? Is this stop-gap fix necessary?

shubhamkumar13:
Are there any specific upsides to adding `num_domains`?

The current fix should handle both the parallel and sequential benchmarks, I essentially added another check which just checked if the json file contained `"runs": []` and if it was there I replaced that with
```
"runs": [
    {
      params: ""
  }
]
```
I don't know if we need `num_domains` as an additional info if a small conditional check can fix it.

Although we can add another target if we want to seperate parallel benchmarks and serial benchmarks

Sudha247:
@shubhamkumar13 replacing `"runs": []` with a run with no parameter might lose any parameters originally specified for the benchmark in the config file.

kayceesrk:
In addition to @Sudha247's comment, having explicit `num_domains` would mean that we don't have to "assume" that the first parameter is `num_domains`. It is explicit in the `json` file. This would be better documentation for external users who wish to contribute code.

kayceesrk:
I'll close this PR. Let's do this properly with `num_domains`.

Issue #138, opened 2020-07-14
Enrich the functionality of .comp files


kayceesrk:
Currently, the compiler variants in `.comp` files only take the URL. We've seen many requests for enriching the functionality of `.comp` files so that the builds and runs may be customized. Some of these are:

1.  `OCAMLRUNPARAM` parameters for the runs. Consider the case of running the same compiler variant with different `OCAMLRUNPARAM` parameters. Currently, there is no way to configure this as part of the variant and must be externally applied during runs. 
2. Compiler configuration flags such as enabling flambda.
3. Different wrappers (orun, pauseimes, perf). These are now hardcoded in the `.json` files [1]. It will be better if these were described in the `.comp` files.

I would consider 1 and 2 to be high priority right now. Once that is done, we can do 3. I recommend using s-expression for the new data format of `.comp` files, and parse it with the help of sexplib [2]. It might also be useful to rename `.comp` to `.var` to signify that these are different variants not compilers (same compiler might have different variants c.f. item 1 above). 

[1] https://github.com/ocaml-bench/sandmark/blob/master/multicore_parallel_run_config.json#L2-L15
[2] https://github.com/janestreet/sexplib

@shakthimaan

kayceesrk:
Fixed by #141.

PR #141, opened 2020-07-15
Enrich the variants with additional options

shakthimaan:
The following PR address the issue #138 to enrich the functionality of .comp files.
# Status
- [X] The .comp file has been renamed as .json file to denote that this is a variant file.
- [X] We use JSON syntax to specify URL, configure options to ocaml-base-compiler and arguments to OCAMLRUNPARAM. The URL is mandatory while you can leave out the configure and runparams entries, or just use empty quotes for their arguments. An example ocaml-versions/4.10.0+multicore.var file is given below:
```
{
  "url" : "https://github.com/ocaml-multicore/ocaml-multicore/archive/parallel_minor_gc.tar.gz",
  "configure" : "-q",
  "runparams" : "v=0x400"
}
```
- [X] Rename all existing .comp files to .json and use the JSON format change.
- [X] Update README.
Note: Support for wrappers (orun, pausetimes, perf) instead of using the .json files. RFC discussion at https://github.com/ocaml-bench/sandmark/issues/142.
# Testing
1. You can run the following to check for `sexp` dependency:
```
$ make depend
```
2. You can build the dependencies, including ocaml-base-compiler, after updating the ocaml-versions/4.10.0+multicore.json file and executing the following command:
```
$ BUILD_ONLY=1 make ocaml-versions/4.10.0+multicore.bench
```
* The configure option can be verified from the `./configure` command output in _opam/4.10.0+multicore/.opam-switch/build/ocaml-base-compiler.4.10.0+multicore/config.log file. 
* You can also verify the actual arguments passed in the output of the above build command. For example: OCAMLRUNPARAM=v=0x400 OCAMLCONFIGOPTION=-q opam switch create --keep-build-dir --yes 4.10.0+multicore ocaml-base-compiler.4.10.0+multicore.


kayceesrk:
> OCAMLRUNPARAM=v=0x400 OCAMLCONFIGOPTION=-q opam switch create --keep-build-dir --yes 4.10.0+multicore ocaml-base-compiler.4.10.0+multicore.

Should you add quotes around the options so that the bash variables are correctly assigned. For example

> OCAMLCONFIGOPTION="-q -disable-naked-pointers"

will work but

> OCAMLCONFIGOPTION=-q -disable-naked-pointers

would not.

ctk21:
This is going to add some much needed features, but I'm wondering if there's an undesirable side effect to this design: it will result in multiple builds of the compiler and opam universe.

For example suppose that the user wanted to run the following experiment against 4.10.0+stock to measure the impact of the minor heap size. The run plan they are after is:
  4.10.0+stock, OCAMLRUNPARAM=s=1M, using orun
  4.10.0+stock, OCAMLRUNPARAM=s=2M, using orun
  4.10.0+stock, OCAMLRUNPARAM=s=4M, using orun
  4.10.0+stock, OCAMLRUNPARAM=s=8M, using orun
  4.10.0+stock, OCAMLRUNPARAM=s=16M, using orun
  4.10.0+stock, OCAMLRUNPARAM=s=32M, using orun

By pushing the OCAMLRUNPARAM (and wrapper) into the opam compiler build config aren't we going to end up building the compiler and opam universe 6 times for the above experiment?

Aside the extra builds, the user is going to need to setup 6 nearly identical `var` files to run that experiment. That may be ok if we assume that users just script things up, but it does push all the config management, execution and output naming back to them.

kayceesrk:
@ctk21 this is an issue that we have currently as our namespacing for variants is tied to the compiler switch. For example, for the ICFP paper, we had 

* [4.06.1+multicore.comp
](https://github.com/ocaml-bench/sandmark/blob/master/ocaml-versions/4.06.1%2Bmulticore.comp)
* [4.06.1+multicore+parallel.comp](https://github.com/ocaml-bench/sandmark/blob/master/ocaml-versions/4.06.1%2Bmulticore%2Bparallel.comp)
* [4.06.1+multicore+pausetimes.comp
](https://github.com/ocaml-bench/sandmark/blob/master/ocaml-versions/4.06.1%2Bmulticore%2Bpausetimes.comp)

all of which builds the same compiler + dependencies + benchmarks.

ctk21:
> We should add an issue to make this situation better. In particular, the notion of variants should be separated from the OPAM switches. Multiple variants all of which build the same compiler + package, but differ by `OCAMLRUNPARAM` or wrappers (perf, pausetime, orun) should share the compiler switch.

Isn't that possible as things stand now? I mean you build the single switch and then alter the wrapper with `RUN_BENCH_TARGET` to give you separate `_results/<variant_name>/<variant_name>.<wrapper_name>.bench` output files.

kayceesrk:
@shakthimaan As agreed in the meeting, if you skip

>  Support for wrappers (orun, pausetimes, perf) instead of using the .json files.

and finish the other two, we can merge this one.

PR #143, opened 2020-07-17
Added Coq library fraplib and a benchmark that depends on it

kayceesrk:
`fraplib` library is the library part of the [FRAP book](https://github.com/achlipala/frap). The build of the library has been dunified and lives [here](https://github.com/kayceesrk/fraplib). This is addition is intended to showcase an example of adding Coq benchmarks into Sandmark. There are a few more long-running lectures/coq sources in the FRAP book. It would make sense to add them here. 

@shubhamkumar13


kayceesrk:
@shakthimaan @ctk21 Any comments on this one?

dune-project
```diff
@@ -1 +1,2 @@
-(lang dune 1.7)
+(lang dune 2.5)
```
shakthimaan, line 1: This can be set to 2.6. The build and benchmarks run fine with this change.

PR #144, opened 2020-07-17
Add frap as a coq benchmark

shubhamkumar13:
Added all the coq files present in [frap](https://github.com/achlipala/frap) as a coq benchmark.


shubhamkumar13:
@kayceesrk I've added all the coq files present in the [frap[(https://github.com/achlipala/frap) repo.

I've tested them locally and they don't seem to show any errors.

Any comments are welcome

kayceesrk:
I don't think adding all the coq files is useful. Remember that benchmarking is a process where we aim to get good coverage of all the potential behaviours of the compiler in _reasonable_ time. Adding all the benchmarks creates a lot of data. But do we know that every one of them is useful? Are each of the runs bringing something interesting or all of the runs look the same? The benchmark runs already take such a long time to finish that we don't run them in the CI (which I still think we should fix; we should at least run a subset of the benchmarks). 

I would recommend doing the following. Run the benchmarks and paste the results for the individual coq runs in this thread. Just the results in `4.10.0+multicore.orun.bench` should suffice. Then we can pick the interesting ones to include.

shubhamkumar13:
This file has considerably high running time
```
real	0m28.023s
user	0m27.840s
sys	0m0.180s
```
Also the number of memory allocations were considerably high as a benchmark
```
allocated_words: 9830680474
minor_words: 9819150819
promoted_words: 154273778
major_words: 165803433
minor_collections: 37478
major_collections: 29
heap_words: 29199360
heap_chunks: 33
top_heap_words: 29199360
compactions: 0
```

kayceesrk:
lgtm.

Issue #145, opened 2020-07-18
BUILD_ONLY needs to exit with error if package installation fails 


shakthimaan:
At present, the BUILD_ONLY environment variable stops before executing the benchmarks. We need a way to exit with an error status if any of the dependency installation packages fail. We use `--best-effort` to ignore any build errors and proceed.

In particular, `frama-c` (https://github.com/ocaml-bench/sandmark/issues/17) needs to be built successfully for this change to be useful with the CI.

kayceesrk:
#147 fixes this problem in a different way. Rather than attempting to fix `frama-c`, #147 takes the easier route of running a subset of benchmarks in the CI.

PR #147, opened 2020-07-20
Add support to associate tags with benchmarks.

kayceesrk:
The current tags are `macro_bench` and `run_in_ci`. The first tag indicates that the benchmark is a macro benchmark. The second tag indicates that the benchmark must be run in the CI. Modify the CI scripts to run the benchmarks tagged as `run_in_ci`. 

The `run_in_ci` tag circumvents #145. Instead of builds only in CI, whose failures are also ignored as some package builds may fail, we only run a subset of the programs in the CI.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -16,7 +16,8 @@ steps:
   - opam install dune.2.6.0
   - export ITER=1
   - export OPAM_DISABLE_SANDBOXING=true
-  - BUILD_ONLY=1 make ocaml-versions/4.10.0+stock.bench
+  - make run_config_ci.json
+  - RUN_CONFIG_JSON=run_config_ci.json make ocaml-versions/4.10.0+stock.bench
 
 ---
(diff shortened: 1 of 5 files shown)


.drone.yml
```diff
@@ -16,7 +16,8 @@ steps:
   - opam install dune.2.6.0
   - export ITER=1
   - export OPAM_DISABLE_SANDBOXING=true
-  - BUILD_ONLY=1 make ocaml-versions/4.10.0+stock.bench
```
shakthimaan, line 19: The BUILD_ONLY variable assignment:
```
BUILD_ONLY ?= 0
```
and its conditional check can be removed from the Makefile:
```
	@{ if [ "$(BUILD_ONLY)" -eq 0 ]; then												\
```

kayceesrk: I’d leave it in as it might be independently useful.

PR #154, opened 2020-07-30
revert paramwrapper changes

shubhamkumar13:
PR to revert changes in the present in #124 because of the following reasons.

> The PARAMWRAPPER is configured for parallel benchmarks, so while generating run_config_macro.json
> - It overwrites some parameters already configured, making the runs empty. So most of the benchmarks are not being executed in a run.
> - The paramwrapper has the taskset and chrt which is meant for parallel benchmarks and not necessary for sequential benchmarks.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,12 +1,4 @@
 # options for running the benchmarks
-#
-# paramwrapper for passing taskset and chrt details
-# When passing arguments to PARAMWRAPPER do follow the syntax
-# if params < {number of cores} then paramwrapper = {range of cores} else paramwrapper = {range of cores}
-# the above statement is a skeleton and can be replaced by the following example
-# Example : if params < 16 then paramwrapper = 2-15 else paramwrapper = 2-15,16-27
-PARAMWRAPPER=
-
 # benchmark build target type:
 #  - buildbench: build all single threaded tests
 BUILD_BENCH_TARGET ?= buildbench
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -173,8 +165,7 @@ bash:
 	@echo "[opam subshell completed]"
 
 %_macro.json: %.json
-	if [ -z "$$PARAMWRAPPER" ]; then PARAMWRAPPER="if params < 16 then paramwrapper = 2-13 else paramwrapper = 2-13,16-27"; else echo "user configured paramwrapper"; fi; \
-	jq --arg PARAMWRAPPER "$${PARAMWRAPPER}" '{wrappers : .wrappers, benchmarks: [.benchmarks | .[] | select(.tags | index("macro_bench") != null)]} | {wrappers: .wrappers, benchmarks: [.benchmarks | .[] | {executable: .executable, name: .name, tags: .tags, runs: [.runs | .[] as $$x | $$x | .params | split(" ") | first as $$n | $$n | length | if . < 3 then try ($$n | tonumber | if . < ($$PARAMWRAPPER | split(" ") | .[3] | tonumber) then $$x | .paramwrapper |= ($$PARAMWRAPPER | split(" ") | "taskset --cpu-list " + .[7] + " chrt -r 1") else $$x | .paramwrapper |= ($$PARAMWRAPPER | split(" ") | "taskset --cpu-list " + .[11] + " chrt -r 1") end) else $$x | .paramwrapper |= ($$PARAMWRAPPER | split(" ") | "taskset --cpu-list " + .[7] + " chrt -r 1") end] } ] }' < $< > $@
+	jq '{wrappers : .wrappers, benchmarks: [.benchmarks | .[] | select(.ismacrobench == true)]}' < $< > $@
```
Sudha247, line 168: The tag was changed to `macro_bench` from `ismacrobench` in the config files, for example: https://github.com/ocaml-bench/sandmark/blob/master/run_config.json#L24. So, I think the condition needs to be `select(.tags | index("macro_bench") != null)`.

shubhamkumar13: Thanks @Sudha247 for pointing that out, I didn't notice that tags were added.

PR #155, opened 2020-08-03
sequential version of evolutionary algorithm

Sudha247:


kayceesrk:
Should we add this to `run_config.json` as well?

Sudha247:
Added this to `run_config.json`. Also, removed DLS usage so that the benchmark builds with stock OCaml.

PR #156, opened 2020-08-03
Fix calculation of nbody_multicore

ctk21:
This PR fixes the calculation of the interactions of the bodies and introduce local ref var for velocities to reduce writes and cache traffic.

Given that the inner loop is iterating over all `j  = 0 to Array.length bodies - 1` we should not update the `b'` velocities; updating these velocities is also not parallel domain safe. 

This PR also cuts the false sharing caused by the mutable writes by introducing local reference variables. 
A quick run with `nbody_multicore.exe 16 512 2048` gave:
 - no ref vars: ~1.87s
 - with ref vars: ~1.32s

### DIFF
--- a/benchmarks/multicore-numerical/nbody.ml
+++ b/benchmarks/multicore-numerical/nbody.ml
@@ -29,10 +29,6 @@ let advance bodies dt =
         b.vx <- b.vx -. dx *. b'.mass *. mag;
         b.vy <- b.vy -. dy *. b'.mass *. mag;
         b.vz <- b.vz -. dz *. b'.mass *. mag;
-
-        b'.vx <- b'.vx +. dx *. b.mass *. mag;
-        b'.vy <- b'.vy +. dy *. b.mass *. mag;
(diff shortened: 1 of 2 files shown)


kayceesrk:
lgtm.

shubhamkumar13:
So I compared the nbody_multicore running time
the current master branch :
|Domains   |Time   |
|---|---|
|1   |29.262   |
|2   |24.106   |
|4   |12.674   |
|8   |6.641   |
|12   |4.992   |
|16   |5.799   |
|20   |5.598   |
|24   |6.005   |

PR156 : 
|Domains   |Time   |
|---|---|
|1   |18.126   |
|2   |9.579   |
|4   |4.717   |
|8   |2.416   |
|12   |1.655   |
|16   |1.659   |
|20   |1.268   |
|24   |1.129   |

The above results suggest PR156 is performing more in tune with what's expected if we compare it to the current branch.

kayceesrk:
Thanks for the numbers @shubhamkumar13!

PR #158, opened 2020-08-04
fix for key error in the python notebook for grammatrix

shubhamkumar13:
The fix solves the issue of a `Key Error` while running the python notebook in `notebooks/parallel/parallel.ipynb`

The problem was the empty `params` in the serial benchmark of the grammatrix which generated `grammatrix.` as the key instead of `grammatrix.16`


kayceesrk:
lgtm!

PR #162, opened 2020-08-20
Avoid installing packages which are unused for multicore runs

kayceesrk:
This is mainly to avoid installing `coq` and `fraplib` when running parallel benchmarking. Building `coq` takes ages, and we don't even use it for the parallel runs. This PR is just a start. Ideally, I'd like to have the `PACKAGES` variable completely and accurately describing the required set of packages for serial and parallel runs so that benchmarking runs can be faster. @shakthimaan @shubhamkumar13.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,20 +29,20 @@ CONTINUE_ON_OPAM_INSTALL_ERROR ?= true
 WRAPPER = $(subst run_,,$(RUN_BENCH_TARGET))
 
 PACKAGES = \
-       cpdf conf-pkg-config conf-zlib bigstringaf decompress camlzip menhirLib menhir \
-       minilight base stdio dune-private-libs dune-configurator camlimages yojson lwt \
-       alt-ergo zarith integers js_of_ocaml-compiler uuidm react ocplib-endian        \
-       nbcodec checkseum sexplib0 irmin-mem cubicle conf-findutils coq fraplib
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -29,21 +29,21 @@ CONTINUE_ON_OPAM_INSTALL_ERROR ?= true
...
 ifeq ($(findstring multibench,$(BUILD_BENCH_TARGET)),multibench)
 	PACKAGES += lockfree kcas domainslib ctypes.0.14.0+multicore
-else ## ctypes and frama-c do not build under multicore
-	PACKAGES += ctypes.0.14.0+stock frama-c
+else
+	PACKAGES += ctypes.0.14.0+stock frama-c coq fraplib alt-ergo js_of_ocaml-compiler
 endif
 
+DEPENDENCIES = libgmp-dev libdw-dev jq python3-pip # Ubuntu
+PIP_DEPENDENCIES = intervaltree
```
shubhamkumar13, line 44: Is adding pip dependencies related to coq or fraplib?

shakthimaan: This is required for a clean build with CI for .drone.yml. It is used in Jupyter notebooks.

shubhamkumar13: This looks good then

kayceesrk: These lines were just moved around for better readability.

kayceesrk:
Thanks!

PR #166, opened 2020-08-24
Graph 500 SEQ

singhalshubh:
- graph500.org benchmark
- sequential algorithm implemented
- Support for disconnected graphs is also added.
- For executing, use compile.sh and run.sh(with 3 parameters, scale, edgefactor and startVertex for shortest path).


kayceesrk:
Hi @singhalshubh. Have a look at how the other benchmarks are organized and executed. See `run_config.json`. We don't use bash scripts to execute the benchmark. Please remove the bash scripts and integrate the benchmark into the Sandmark build workflow. I would also recommend going through the README file to understand how we run the benchmarks.

PR #169, opened 2020-08-25
Add check_url for .json file and pkg-config,m4 in Makefile

shakthimaan:
This PR:
* Adds a `check_url` dependency in the Makefile that goes through all the `ocaml-versions/*.json` files to ensure that a URL parameter exists. If not found, it prints the filename.
* On a pristine Ubuntu installation, the `pkg-config` and `m4` packages are required to be installed for Sandmark, and hence the same have been included to `DEPENDENCIES`.


kayceesrk:
Nice!

PR #172, opened 2020-09-08
Remove Base, Stdio orun dependency for trunk

shakthimaan:
The `orun` dependency for `Base` and `Stdio` is removed using this PR to facilitate building Sandmark with trunk.

### DIFF
--- a/orun/dune
+++ b/orun/dune
@@ -12,7 +12,7 @@
 (executable
  (name orun)
  (public_name orun)
- (libraries base stdio re str cmdliner yojson unix wait4 profiler))
+ (libraries base re str cmdliner yojson unix wait4 profiler))
 
 (rule
  (targets profiler_library_flags.sexp)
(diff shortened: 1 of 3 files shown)


orun/dune
```diff
@@ -12,7 +12,7 @@
 (executable
  (name orun)
  (public_name orun)
- (libraries base stdio re str cmdliner yojson unix wait4 profiler))
```
kayceesrk, line 15: Do you need base?

shakthimaan: Yes. `Base.List`, `Base.Int` and `Base.String` are used in `orun.ml`.

kayceesrk: @sadiqj Does it matter whether `orun` depends on `base` and `stdio`? Being an `orun` dependency shouldn't prevent us from building ocaml trunk in sandmark. I must be missing something.

shakthimaan: The need for `Base` has now been removed from the PR.

sadiqj:
Just checking, are there any blockers to merging this?

PR #174, opened 2020-09-18
Cleanup our use of sudo for chrt

ctk21:
This PR does a cleanup of using sudo with parallel tests. This caused me some pain recently when running big test instances (which take a long time O(hrs) to run) that failed because files were turning up with root permissions in strange places. 

The changes are:
 - remove sudo from the makefile; people should use the capability (or any other alternative) to allow them to run chrt
 - make using chrt an explicit wrapper target for `RUN_BENCH_TARGET`. We wanted to be able to benchmark with and without the chrt command.
 - be a bit more upfront in the console when the benchmarks run and the parameters being used (this is helpful for things that run for a long time, just in case the user has made a mistake)
 - document a method to set permissions for chrt on Linux and other changes to be in line with the above

Hopefully this will save others time in the future, although I'd be interested if there is a use case that people use which I've unexpectedly broken.


kayceesrk:
This looks good to me. Definitely better than the current workflow.

Issue #176, opened 2020-09-28
Include raw baseline observation in the normalized graphs


kayceesrk:
The normalized graphs in sandmark currently do not include the raw baseline observations. For example,

![image](https://user-images.githubusercontent.com/410484/94424757-607efa80-01a8-11eb-9843-ef2a99fe6fa1.png)

The graphs shows the normalized running time comparing two compiler variants. This graph does not convey how much time the baseline actually took. We've had few examples where we've seen 20% speedup / slowdown that can be explained by the fact that the benchmark only runs for a few milliseconds and the difference that we see is due to expected execution time variance and noise. 

It would be useful to have the baseline time included in the normalized graphs as additional information with the benchmark name. For example,

![image](https://user-images.githubusercontent.com/410484/94424975-bc498380-01a8-11eb-8c7e-bffda30c3021.png)

The numbers in the parenthesis indicates the time in seconds for the baseline runs. 

It would be useful to include the raw baseline observation in every normalised graph.

PR #177, opened 2020-09-30
Display raw basline numbers in normalized graphs

shubhamkumar13:
Resolving Issue #176

This change adds raw baseline numbers to the normalized graphs in sequential notebooks


kayceesrk:
Would it be possible to post a screenshot of how the graph looks now?

kayceesrk:
I took a look at the PR. I have a couple of questions. 

Why do you need the argument "display_name" at all? I see that it is the same everywhere. What purpose does it serve?

image

It appears that we always get the baseline time irrespective of what metric we're interested in (time, major_cycles, minor_cycles, etc). We want the baseline to be the metric that we're plotting. 

image

It would be nice to include a graph title which says what the baseline unit is. Time in seconds, or kilobytes, count, or...

shubhamkumar13:
The idea was to create a different column that contains the display name of the column, `display_name = name + raw_baseline_number`. And pass that column to `catplot` function which displays the labels of the bar graph.
Even now the default parameters to `x_axis_disp` is `name`. Which was the case prior to this PR.
Screenshot from 2020-10-02 16-34-57
The function that generates the plot takes the `x_axis_disp` param. Which may or may not be the name parameter, depends upon what you pass to `x_axis_disp` argument.
Screenshot from 2020-10-02 16-33-21

One thing that was my fault was not realizing that the baseline metrics would be different for different normalized graphs. That would remove `display_name` to be passed to every_function, which looks redundant now.

shubhamkumar13:
I've made changes to reflect the baseline numbers for that metric.
For eg.
I need to pass the metric that I need to display (here gc.top_heap_words)
Screenshot from 2020-10-02 17-23-17
And the baseline numbers in the graph correspond to the same metric
gc_top_heap_words

Similarly, if I were to get a normalized graph for maxrss_kB, it would be something like this.
maxrss_kB

The only change is last argument passed to `add_display_name`, something like this : 
Screenshot from 2020-10-02 17-24-22

shubhamkumar13:
The above code had some redundancy (since `add_display_name` and `normalise` accept same arguments)
I moved them inside `normalise`.

Screenshot from 2020-10-02 18-00-51

This removes the need to explicitly create a new `df` everytime which is implied when we are generating a normalized graph.

kayceesrk:
> The above code had some redundancy (since add_display_name and normalise accept same arguments). I moved them inside normalise.

Shouldn't the 3rd argument to `add_display_name` be `topic`?

shubhamkumar13:
Oh that's bad! that was an overlook on my part.

Not being able to see sane diffs might be a result of this, but I've made the changes, thanks for the help!

So now the top heap words
normalized_gc top_heap_words
and time_secs have their respective baseline
normalization_time_secs

kayceesrk:
Can you post the screenshot for `maxrsskb`? 

I think Shakti wanted the PR to be made against the sequential-interactive notebook. Can we do that in addition to this change @shubhamkumar13?

shubhamkumar13:
This is how the normalized `maxrsskb` graph looks 
max_rss_raw_baseline

kayceesrk:
Thanks!

PR #181, opened 2020-10-06
lock-free map bench

anmolsahoo25:
This patch implements a lock-free map using tries. The design is inspired by this paper - [Link](http://aleksandar-prokopec.com/resources/docs/ctries-snapshot.pdf). 

This benchmark is a decent mix between domains performing work, synchronization via CAS and garbage collection. Preliminary results on my laptop seemed to produce speedups, but testing on IITM machines is producing a slowdown, as cores are added. Debugging before merge.


anmolsahoo25:
Ah, made a mistake in benchmarking. Was making each domain perform a fixed number of operations. Rather, need to split fixed number of operations between domains. Some results - 

Duration
index

Speedup
index2

benchmarks/multicore-lockfree/test_ctrie.ml
```diff
@@ -0,0 +1,25 @@
...
+let insert_percent = int_of_string Sys.argv.(2)
+let num_opers = int_of_string Sys.argv.(3) / threads
+
+let () = Random.init 42
+
+let tree = Ctrie.empty ()
+
+let work () =
+  for _ = 1 to num_opers do
+    if Random.int 100 > insert_percent then
```
ctk21, line 11: You need to use the Domain Local Storage (DLS) to make sure that `Random.int` is safe when used with multiple domains. When using DLS with `Random` scaling may be better because there will be less write contention. 

Take a look at `benchmarks/multicore-numerical/LU_decomposition_multicore.ml` for an example that uses DLS to setup random state per domain.

anmolsahoo25: Thank you @ctk21 ! That's solid advice. I will try it out.

anmolsahoo25:
After implementing DLS, results look much better. Still some pathology around 20 cores though.
index
index2

benchmarks/multicore-lockfree/test_ctrie.ml
```diff
@@ -0,0 +1,27 @@
+let threads = int_of_string Sys.argv.(1)
+let insert_percent = int_of_string Sys.argv.(2)
+let num_opers = int_of_string Sys.argv.(3) / threads
```
ctk21, line 3: There will be settings of `Sys.argv.(3)` and `threads` where the total number of operations is not the same between runs. 
e.g. `Sys.argv.(3) = 100` and `threads=128` 

Better to use `parallel_for` in domainslib.

benchmarks/multicore-lockfree/test_ctrie.ml
```diff
@@ -0,0 +1,27 @@
...
+      Ctrie.insert (rand_int 1000000) 0 tree 
+  done
+
+let () =
+    let rec spawn_thread n =
+        match n with
+        | 0 -> []
+        | _ -> (Domain.spawn work) :: spawn_thread (n-1)
+    in
+    let threads = spawn_thread (threads - 1) in
```
ctk21, line 25: Might be better to use domainslib `parallel_for` if you want scaling beyond 16-32 cores.

ctk21 (commented):
I haven't spent quality time reviewing the underlying data structure (which looks interesting and worthwhile).

However there are a couple of problems with the work distribution; these could be easily fixed by using domainslib's `parallel_for`.

I would also recommend increasing the benchmark size; looking at scaling up above 16 domains when the runtime is already 0.7s might get quite noise. 

I'd be interested if that improves the scaling you see.

anmolsahoo25:
Thank you for the review. I will get to it this weekend and resolve the issues, so that we can hopefully add it to the sandmark testsuite.

Sudha247:
Hey @anmolsahoo25, [lockfree](https://github.com/ocaml-multicore/lockfree) might be a good place for the lockfree map. Feel free to submit a patch!

anmolsahoo25:
Thanks @Sudha247 , will do! :D

Issue #182, opened 2020-10-08
Rename (or remove?) lu_decomposition benchmark in numerical-analysis


sadiqj:
We have `LU_decomposition` in `multicore-numerical` which can be a source of confusion.

If the `numerical-analysis` implementation doesn't differ significantly from the one in `multicore-numerical` then we should just remove it - otherwise we should rename it.

ctk21:
The implementations are different. Maybe a rename is easiest? 
Say `numerical-analysis/lu_decomposition.ml` -> `numerical_analysis/crout_decomposition.ml`

ctk21:
Fixed by #183

PR #183, opened 2020-10-13
Use crout_decomposition name for numerical analysis benchmark

shakthimaan:
Renaming `lu_decomposition` to `crout_decomposition` to avoid confusion with `LU_decomposition` in `multicore-numerical`. The implementations are different, and hence the rename as suggested in https://github.com/ocaml-bench/sandmark/issues/182.


ctk21:
LGTM. Fixes #182

PR #186, opened 2020-10-21
Very simple effect microbenchmarks to cover code paths

ctk21:
This PR introduces very simple micro benchmarks that are looking at the throughput of our effects system.

In particular it has 4 tests:
 - `effect_throughput_clone`: this captures a continuation and then does `continue (Obj.clone_continuation k) x` on it. 
 - `effect_throughput_val`: this executes a function inside an effect handler, but returns without performing an effect. 
 - `effect_throughput_perform`: this executes a function inside an effect handler, performs an effect, which resumes the continuation which exits with a value.
 - `effect_throughput_perform_drop`: this executes a function inside an effect handler, performs an effect, which drops the continuation and so the exit is from the effect handler itself.

These tests are not exhaustive, but are intended to give code coverage to flush out any gross inefficiencies and can be useful for CI/debugging purposes. 

I'm open to more or better versions of the above.

### DIFF
--- a/benchmarks/multicore-effects/dune
+++ b/benchmarks/multicore-effects/dune
@@ -1,3 +1,21 @@
+
+
+(executables
+    (names effect_throughput_clone)
+    (modules effect_throughput_clone))
+
+(executables
+    (names effect_throughput_perform)
+    (modules effect_throughput_perform))
+
+(executables
+    (names effect_throughput_perform_drop)
+    (modules effect_throughput_perform_drop))
+
+(executables
+    (names effect_throughput_val)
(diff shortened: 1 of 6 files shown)


kayceesrk:
The benchmarks are quite useful. I'll merge this PR.

We should also consider effect handler version of highly recursive programs such as recursive Fibonacci, tak, and ackermann's function.

ctk21:
> We should also consider effect handler version of highly recursive programs such as recursive Fibonacci, tak, and ackermann's function.

I had these waiting in the wings: https://github.com/ocaml-bench/sandmark/pull/187

PR #188, opened 2020-10-23
Issue Solved, Classification of benchmarks.

singhalshubh:
[SEQ] -  Classification of benchmarks
- Updated run_config.json
- Classified into "lt_1s", 1s_10s", "10s_100s",  "gt_100s" as 4 categories.


ctk21:
This PR is removing the old `tags` and replacing them with just the time attributes.

Given that `tags` is a list of free strings that people can use to filter on, is it possible to add the time attributes but leave the existing `tags`? For example with `cpdf`, I would have expected `tags` to become:
```json
"tags": [
        "macro_bench",
        "run_in_ci",
        "10s_100s"
      ]
```

singhalshubh:
> This PR is removing the old `tags` and replacing them with just the time attributes.
> 
> Given that `tags` is a list of free strings that people can use to filter on, is it possible to add the time attributes but leave the existing `tags`? For example with `cpdf`, I would have expected `tags` to become:
> 
> ```json
> "tags": [
>         "macro_bench",
>         "run_in_ci",
>         "10s_100s"
>       ]
> ```

Hey, I have updated the tags.

singhalshubh:
Updated Multicore too.

singhalshubh:
Micro-Multicore done.

Issue #184, opened 2020-10-16
Running benchmarks with varying OCAMLRUNPARAM


ctk21:
Right now there isn't an easy way to run experiments where OCAMLRUNPARAM changes but the compiler build stays fixed. 

Suppose we wanted to look at the impact of the minor heap size across our benchmarks with 4.10.0+stock.

The run plan we want is:
4.10.0+stock, OCAMLRUNPARAM=s=1M, using orun
4.10.0+stock, OCAMLRUNPARAM=s=2M, using orun
4.10.0+stock, OCAMLRUNPARAM=s=4M, using orun
4.10.0+stock, OCAMLRUNPARAM=s=8M, using orun
4.10.0+stock, OCAMLRUNPARAM=s=16M, using orun
4.10.0+stock, OCAMLRUNPARAM=s=32M, using orun

One way to do this might be to have a way to specify `OCAMLRUNPARAM` within a wrapper in the `run_config.json`; this has the advantage that we get the wrapper naming and reuse of the build for free. There may be other better ways to do it. 

(NB: the `runparams` in the compiler variant [here](https://github.com/ocaml-bench/sandmark#configuration) is only applied to the opam build, not the run of the benchmarks)

ctk21:
So another use case which is a bit more general, but I think we should include in any fix here, is being able to set arbitrary environment variables for the benchmark run. 

For example, an experiment people wanted to be able to run recently was glibc malloc vs mimalloc vs jemalloc vs tcmalloc vs etc... In theory it might be as simple as specifying `LD_PRELOAD` in the environment, which if we can somehow just set from within the `run_config.json` then that would be good.

shakthimaan:
Does the following syntax in an `ocaml-versions/4.10.0+stock.json` file be sufficient?
```
{
  "url" : "https://github.com/ocaml/ocaml/archive/4.10.0.tar.gz",
  "configure" : "-q",
  "runparams" : "v=0x400",
  "experiments" : {
      "test1" : "OCAMLRUNPARAM='s=1M'",
      "test2" : "OCAMLRUNPARAM='s=2M'",
      "test3" : "OCAMLRUNPARAM='s=4M'",
      "test4" : "OCAMLRUNPARAM='s=8M'"
      "testj" : "LD_PRELOAD=/path/to/jemalloc.so OCAMLRUNPARAM='v=0x400'"
   }
}
```
The output will be a different _results (`test1`, `test2`, etc.) folder for each experiment.

[...]

PR #189, opened 2020-10-26
Add environment support for wrapper in JSON configuration file

shakthimaan:
The following PR implements `Option A` as mentioned in https://github.com/ocaml-bench/sandmark/issues/184. An (optional) `environment` parameter can specified for a wrapper entry in `run_config*.json` as shown below:
```
    {
      "name": "orun_2M",
      "environment": "OCAMLRUNPARAM='s=2M'",
      "command": "orun -o %{output} -- taskset --cpu-list 5 %{command}"
    },

```
* You can simply add multiple `variable=values` entries to the `environment` parameter value, and it will be passed during run-time execution.
* The `subst` in the Makefile to extract the wrapper name has been replaced by `patsubst` to allow bench target names with underscore like `run_orun_2M`.

You can run the above using:
```
$ RUN_BENCH_TARGET=run_orun_2M make ocaml-versions/4.10.0+multicore.bench
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -89,7 +89,6 @@ _opam/%: _opam/opam-init/init.sh ocaml-versions/%.json setup_sys_dune
 	  >> dependencies/packages/ocaml-base-compiler/ocaml-base-compiler.$*/opam
 	$(eval OCAML_CONFIG_OPTION = $(shell jq -r '.configure // empty' ocaml-versions/$*.json))
 	$(eval OCAML_RUN_PARAM     = $(shell jq -r '.runparams // empty' ocaml-versions/$*.json))
-	$(eval ENVIRONMENT = $(shell jq -r '.wrappers[] | select(.name=="$(WRAPPER)") | .environment // empty' "$(RUN_CONFIG_JSON)" ))
 	opam update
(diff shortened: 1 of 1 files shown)


Engil:
Thank you very much for this!
This PR looks good to me, I gave it a few trial runs.
One missing bit would be to fix the notebooks so that it does not rely on `orun` being in the name of a given benchmark result, but rather use a regexp to glob against `orun_*` since it looks like this is the pattern we are going for.
Maybe this can be addressed in another PR, as far as this one goes I think it is good.

shakthimaan:
> One missing bit would be to fix the notebooks so that it does not rely on `orun` being in the name of a given benchmark result, but rather use a regexp to glob against `orun_*` since it looks like this is the pattern we are going for.
> Maybe this can be addressed in another PR, as far as this one goes I think it is good.

Yes, the following code snippet needs to be used. Will update the notebook PRs shortly:
```
for root, dirs, files in os.walk(artifacts_dir):
    for file in files:
        if file.endswith((".orun.bench", ".orunchrt.bench")):
...
```

ctk21:
merging to get the functionality

PR #192, opened 2020-11-06
GraphSEQ Corrected

singhalshubh:
Earlier, there were concerns regarding the run time of the merged graph500. There was an error in Kronecker generator which has now been rectified.
(For scale = 13, it gives nice > 70s results)

Timings of benchmarks : 
Kernel1 : 1m10.058s
Kernel 2 :  5.041s
Kernel3 :   1m25.861s
On Turing for scale = 12, and edgefcator = 10

Witness the graph in kronecker.txt
"Kronecker : 1m9.528s" time.


shubhamkumar13:
I see that the [kronecker.ml](https://github.com/ocaml-bench/sandmark/blob/30bdaf424aa92ef81edbb72e84ea058243acbded/benchmarks/graph500seq/kronecker.ml) file and [kernel1.ml](https://github.com/ocaml-bench/sandmark/blob/30bdaf424aa92ef81edbb72e84ea058243acbded/benchmarks/graph500seq/kernel1.ml) uses Lists while [kernel2.ml](https://github.com/ocaml-bench/sandmark/blob/30bdaf424aa92ef81edbb72e84ea058243acbded/benchmarks/graph500seq/kernel2.ml) and [kernel3.ml](https://github.com/ocaml-bench/sandmark/blob/30bdaf424aa92ef81edbb72e84ea058243acbded/benchmarks/graph500seq/kernel3.ml) uses Arrays is there a specific reason why they are using different structures?

benchmarks/graph500seq/kernel1.ml
```diff
@@ -1,17 +1,17 @@
...
+from (startVertex, endVertex, weight) in column to (startVertex, endVertex, weight) in 3 rows*)
 
 let scale = try int_of_string Sys.argv.(1) with _ -> 2
 
 let edgefactor = try int_of_string Sys.argv.(2) with _ -> 1
 
-(* This function helps in transpose of the list which has to be converted from
- * [(startVertex, endVertex, weight)] in column to [(startVertex, endVertex,
- * weight)] in 3 rows *)
 let rec transpose list col newList =
```
shubhamkumar13, line 15: `list` is a type name, try using `lst` instead, also this is a style thing in ocaml that we use snake case and not camel case although that should be very intutive :)

benchmarks/graph500seq/kronecker.ml
```diff
@@ -92,33 +96,40 @@ let rec printList list =
...
-  let _ = printList ijw in
   let ijw = permute (transpose 0 ijw []) in
   let ijw = transpose 0 ijw [] in
+  let _ = printList ijw in
+  if Sys.file_exists "kronecker.txt" then Sys.remove "kronecker.txt"; 
+  let file = open_out "kronecker.txt" in
+  let _ = writeFile ijw file in
+  let _ = close_out file in
   ijw
+;;
```
shubhamkumar13, line 133: `;;` can be removed

benchmarks/graph500seq/kronecker.ml
```diff
@@ -92,33 +96,40 @@ let rec printList list =
...
+
+kronecker scale edgefactor;;
```
shubhamkumar13, line 135: We rarely use `;;` in source code ([why?](https://baturin.org/docs/ocaml-faq/))
To run this idiomatically we usually use `let () = kronecker ..`

benchmarks/graph500seq/kronecker.ml
```diff
@@ -92,33 +96,40 @@ let rec printList list =
...
 
 let computeNumber scale edgefactor =
   let n = int_of_float (2. ** float_of_int scale) in
   let m = edgefactor * n in
   (n, m)
 
+let rec writeFile ijw file = 
+  match ijw with 
+  [] -> None |
+  head::tail -> let _ = List.iter (Printf.fprintf file "%f, ") head in let _ = Printf.fprintf file "\n" in writeFile tail file
```
shubhamkumar13, line 110: `let _ = <some code> in let _ = <some other code> in` can be replaced by `<some code> ; <some other code>`
or `<some code> |> fun _ -> <some other code>`, there is nothing wrong with it, just the way ocaml idiomatic code looks

singhalshubh:
> I see that the [kronecker.ml](https://github.com/ocaml-bench/sandmark/blob/30bdaf424aa92ef81edbb72e84ea058243acbded/benchmarks/graph500seq/kronecker.ml) file and [kernel1.ml](https://github.com/ocaml-bench/sandmark/blob/30bdaf424aa92ef81edbb72e84ea058243acbded/benchmarks/graph500seq/kernel1.ml) uses Lists while [kernel2.ml](https://github.com/ocaml-bench/sandmark/blob/30bdaf424aa92ef81edbb72e84ea058243acbded/benchmarks/graph500seq/kernel2.ml) and [kernel3.ml](https://github.com/ocaml-bench/sandmark/blob/30bdaf424aa92ef81edbb72e84ea058243acbded/benchmarks/graph500seq/kernel3.ml) uses Arrays is there a specific reason why they are using different structures?

It was mentioned in kronecker in graph500 bench that we have to use lists. So kronecker generates a list and kernel1 uses that list to make a HashMap (for sparse representation).

benchmarks/graph500seq/kronecker.ml
singhalshubh: I will make the changes.

shubhamkumar13:
I think this PR looks good to merge except kronecker.txt is generated manually right now by running kronecker.exe with the necessary parameters and if one wants to run kernel2 and kernel3, kronecker.txt is a requirement. 

Since kronecker is not a part of the benchmark this becomes a little bit tricky since the .exe file is required, @shakthimaan is doing this manually the right way forward?

shakthimaan:
The kronecker.txt file size is 1.3 MB, and it will be better to generate the file before running the benchmark. Perhaps, it can be built after kronecker.exe is built, and kept in /tmp? It needs to be removed afterwards though.

shubhamkumar13:
So we do need kronecker.exe once and we can use it multiple times. But it also takes input arguments to generate the text file and my only question is do we need to make sure that the arguments passed to `kernel2.exe` and `kernel3.exe` should be similar to what gets passed on to `kronecker.exe` since there are 2 test cases which are passed on as arguments to `kernel2` and `kernel3`. 

This is might be the only reason to rebuild it @singhalshubh do we need to match the arguments passed on to `kronecker` with `kernel2` and `kernel3`?

singhalshubh:
Yes. Kernels and kronecker should have the same arguments. Kernels might have additional arguments.

shubhamkumar13:
@shakthimaan I think this answers your question. As we need to have the same arguments across all the files for every test case the kronecker.txt files need to be rebuilt.

kayceesrk:
What more work is needed on this PR?

shakthimaan:
The kronecker.txt is 1.3 MB, and hence, it was suggested to generate the same before the benchmarks are executed. Can anyone verify the same? Otherwise, the PR is ready to be merged.

singhalshubh:
> The kronecker.txt is 1.3 MB, and hence, it was suggested to generate the same before the benchmarks are executed. Can anyone verify the same? Otherwise, the PR is ready to be merged.

I have already made the required changes on Wed after we spoke. Is there any more work?

shakthimaan:
@shubhamkumar13 Kindly verify the latest changes, and the PR can be merged. Thanks!

shubhamkumar13:
@shakthimaan @singhalshubh I was able to generate the kronecker.txt file manually after running the benchmark and I used a similar manual process to run `kernel1`, `kernel2` and `kernel3`. If this is how it's supposed to be done then it works fine and maybe adding a note to the readme or maybe inside the `graph500seq` should be a nice way to document the process to use the benchmark. Apart from that I think this can be merged

singhalshubh:
> @shakthimaan @singhalshubh I was able to generate the kronecker.txt file manually after running the benchmark and I used a similar manual process to run `kernel1`, `kernel2` and `kernel3`. If this is how it's supposed to be done then it works fine and maybe adding a note to the readme or maybe inside the `graph500seq` should be a nice way to document the usage. Apart from that I think this can be merged

It doesnt need a manual generation anymore. Kronecker would develop .txt file without any manual interference.

shubhamkumar13:
>It doesnt need a manual generation anymore. Kronecker would develop .txt file without any manual interference.
Yes it does, sorry that I didn't see this earlier @shakthimaan I think this PR can be merged then.

PR #194, opened 2020-11-18
Lockfree benchmarks

Sudha247:
This patch contains benchmarks for existing structures in the [lockfree](https://github.com/ocaml-multicore/lockfree) library. All lockfree benchmarks are labelled with a `lockfree_bench` tag.

Benchmark for a data structure contains a sequential and parallel version. Parallel version splits the available work equally to all domains.


Sudha247:
Some results for these benchmarks:
**Time**

image

**Speedup**

image

kayceesrk:
This looks useful. Are the non-scaling benchmarks expected not to scale?

Sudha247:
While I wouldn't expect linear scaling in the non-scaling benchmarks, the current numbers need investigation.

ctk21:
One thing with these Lockfree benchmarks is that the 'time' and 'speedup' don't tie out with how many of the papers in the area are reporting their numbers. Is it possible to also report the numbers in terms of throughput (e.g. number of transactions per second) or in terms of time per transaction?

kayceesrk:
I assume that these benchmarks don't have much parallelisable work. For example, consider the benchmark for lock-free queues with a producer and consumer pushing items through the queue. It is unclear whether to expect speedup by splitting the items among multiple producers and consumers, but using the same shared queue. There is increased contention on the shared queue, and there isn't any useful parallelisable work. 

Given this, adding some commentary in the top of the test program files would help someone looking at these numbers in the future. I am happy to merge this PR as is, but let's open an issue to track the task to add commentary to these benchmarks. Is this acceptable @Sudha247?

Sudha247:
Yeah, that sounds good! Like Tom mentioned above, it might be useful to measure throughput vs number of cores rather than speedup. https://github.com/ocaml-multicore/lockfree/issues/6 has some details about these benchmarks, perhaps we could open a placeholder issue on sandmark too.

kayceesrk:
@Sudha247 sounds good. Can you open the placeholder issues? I'll merge this PR now.

Issue #195, opened 2020-11-19
Filtering benchmarks based on tags


Sudha247:
We have multiple tags to classify benchmarks in the config files, like `macro_bench`, `run_in_ci` and tags based on running time. Makefile currently supports filtering macro and CI benchmarks with custom rules. 

It would be useful to generalize this further by taking the following items as inputs:
* tag
* source file
* destination file

All benchmarks with the `tag` in source file needs to be copied to destination file. 

The following `jq` filter could be used to do the filtering:

```
jq '{wrappers : .wrappers, benchmarks: [.benchmarks | .[] | select(.tags | index(<tag>) != null)]}'
```

kayceesrk:
We could build this into the makefile as

```
$ make run_config TAGS="macro_bench"
```

and have a target called `run_config` which uses the environment variable `TAGS` to filter those tags. 

Is there a way to accept more than one tag in that command?

kayceesrk:
> Is there a way to accept more than one tag in that command?

Perhaps this is not a good idea. It is unclear whether the filter with multiple tags should act as a conjunction or disjunction.

kayceesrk:
See https://github.com/ocaml-bench/sandmark/pull/196.

shakthimaan:
> Is there a way to accept more than one tag in that command?
This is a useful feature request, considering that we would like to classify benchmarks based on the family of algorithms as well. A researcher might be interested to run those benchmarks that are macro_benchmarks, for example, AND finish in less than 10s AND belong to graph algorithms.

kayceesrk:
We can do conjunction by filtering the filtered file. See https://github.com/ocaml-bench/sandmark/pull/196/files#diff-b54b39f1afced2465e1f3641db9d5bbf4f3a7fcf890996dfedd3c197bcb7f8c7R66-R67. Disjunction is trickier. 

Let's review and merge #196 first, given that we can already do conjunction, albeit externally.

kayceesrk:
> A researcher might be interested to run those benchmarks

Let's also wait until this researcher turns up :-)

PR #196, opened 2020-12-02
Filter benchmarks based on tag

kayceesrk:
Remove the custom targets `_macro.json` and `_ci.json` and replace that with a generic solution. See #195.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -17,8 +17,8 @@ steps:
   - opam install dune.2.6.0
   - export ITER=1
   - export OPAM_DISABLE_SANDBOXING=true
-  - make run_config_ci.json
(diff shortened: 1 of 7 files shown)


kayceesrk:
@Sudha247 please see if this is what you were looking for.

Sudha247:
Thanks for this solution @kayceesrk! This is what I was looking for. I have only one suggestion; It might useful to remove `*_filtered.json` during `make clean`. A note on this feature (in the README perhaps) would be helpful, too. :)

shakthimaan (approved):
LGTM

kayceesrk:
> I have only one suggestion; It might useful to remove *_filtered.json during make clean. A note on this feature (in the README perhaps) would be helpful, too. :)

Done.

PR #202, opened 2021-01-15
Added bench clean target in the Makefile

shakthimaan:
Add a `benchclean` target to remove the generated benchmarks and the benchmark results.


ctk21:
LGTM, merging!

PR #203, opened 2021-01-25
Implement ITER support

shakthimaan:
The following patch implements ITER support when building and executing benchmarks. When used with ITER=2 for 4.10.0+multicore (for example), it correctly creates two separate directories for the benchmarks:
```
sandmark $ ls _build/
4.10.0+multicore_1  4.10.0+multicore_2  log
```
The final consolidated .bench file now uses the ITER value:
```
$ ls _results/
4.10.0+multicore_1.orun.summary.bench  4.10.0+multicore_2.orun.summary.bench
```
You can test the patch using:
```
$ TAG='"run_in_ci"' make run_config_filtered.json
$ ITER=2 RUN_CONFIG_JSON=run_config_filtered.json make ocaml-versions/4.10.0+multicore.bench 
```


ctk21:
Looks reasonable to me, merging!

PR #204, opened 2021-01-26
Adding layers.ml as a benchmark to sandmark

shubhamkumar13:
This PR adds the layers benchmark to the sandmark suite. The layers benchmark is one of the benchmarks present in the [mirage/irmin/bench/irmin-pack](https://github.com/mirage/irmin/blob/master/bench/irmin-pack/layers.ml) repo. 

The benchmark (as far as I can tell) creates a store of type [`Irmin_pack_layered`](https://mirage.github.io/irmin/irmin-pack/Irmin_pack_layered/index.html) and generates a tree structure for that layered store, checks out the branch of the tree and commits to the store. The stats for this complete cycle is calculated (for each commit) and it prints out those stats after the run.

The current PR adds the layers benchmark and it's dependencies to sandmark. The layers benchmark contains a lot of ppx code so the layers.ml benchmark file added to sandmark contains ppx generated code to avoid preprocessing on every sandmark run. 

But we do require ppx libraries and one of the major reasons we had to [modify](https://github.com/shubhamkumar13/sandmark/blob/d89647d82c1449b3b75de93584183f93b3c3b03c/Makefile#L115) the Makefile for ppx support. Ppx libraries require omp and ppxlib so we had to create a conditonal check for different variants of ocaml (multicore/stock/domains/domains+effects) to install `ocaml-migrate-parsetree` and `ppxlib` which compile with that ocaml variant. This work was only possible because @Sudha247 was able to create `ocaml-migrate-parsetree` and `ppxlib` versions which compiled with the multicore variant of ocaml.


Makefile
```diff
@@ -107,10 +108,18 @@ blah:
...
+		opam pin add -n --yes --switch $* ppxlib.dev https://github.com/Sudha247/ppxlib.git#multicore; 					 \
+		echo "Multicore" ;;  \
+	  *) echo "Not mutlicore";;  \
+	  esac };
 	$(eval ENVIRONMENT = $(shell jq -r '.wrappers[] | select(.name=="$(WRAPPER)") | .environment // empty' "$(RUN_CONFIG_JSON)" ))
 	@opam update
 	opam install --switch=$* --keep-build-dir --yes rungen orun
 	opam install --switch=$* --best-effort --keep-build-dir --yes $(PACKAGES) || $(CONTINUE_ON_OPAM_INSTALL_ERROR)
+	opam exec --switch $* -- opam list
+	opam list --rec --resolve irmin
```
shubhamkumar13, line 122: Remove uneccessary commands

benchmarks/irmin/dune-project
```diff
@@ -0,0 +1 @@
+(lang dune 2.8)
```
shubhamkumar13, line 1: This file should be deleted, it was produced as a side effect of local testing

Makefile
```diff
@@ -107,10 +108,18 @@ blah:
 	@echo ${PACKAGES}
 
 ocaml-versions/%.bench: check_url depend log_sandmark_hash ocaml-versions/%.json _opam/% .FORCE
+	@{ case "$*" in \
+	  *multicore*) opam pin add -n --yes --switch $* ocaml-migrate-parsetree.dev https://github.com/Sudha247/ocaml-migrate-parsetree.git#multicore; \
+		opam pin add -n --yes --switch $* ppxlib.dev https://github.com/Sudha247/ppxlib.git#multicore; 					 \
```
Sudha247, line 113: Can we instead add the `opam` files and extra multicore patch files of these libraries to `dependencies/packages/` directory? Patches are in https://github.com/ocaml-multicore/multicore-opam/pull/46 and https://github.com/ocaml-multicore/multicore-opam/pull/47. It is easy to miss seeing that the patches are being pinned here for multicore compiler versions.

Makefile
```diff
@@ -107,10 +108,18 @@ blah:
...
+		echo "Multicore" ;;  \
+	  *) echo "Not mutlicore";;  \
+	  esac };
 	$(eval ENVIRONMENT = $(shell jq -r '.wrappers[] | select(.name=="$(WRAPPER)") | .environment // empty' "$(RUN_CONFIG_JSON)" ))
 	@opam update
 	opam install --switch=$* --keep-build-dir --yes rungen orun
 	opam install --switch=$* --best-effort --keep-build-dir --yes $(PACKAGES) || $(CONTINUE_ON_OPAM_INSTALL_ERROR)
+	opam exec --switch $* -- opam list
```
shakthimaan, line 121: The `opam exec --switch $* -- opam list` is fine, if we later would like to see if all the required dependencies got installed.

shubhamkumar13:
@shakthimaan if this seems fine to merge lets merge it

Makefile
```diff
@@ -108,11 +108,16 @@ blah:
...
+	{ case "$*" in \
+		*multicore*) PACKAGES += ocaml-migrate-parsetree.2.1.0+multicore ppxlib.0.22.0+multicore; echo "Multicore";; \
+	*) PACKAGES += ocaml-migrate-parsetree.2.1.0+stock ppxlib.0.22.0+stock; echo "Not Multicore";; \
+	esac };
 	$(eval ENVIRONMENT = $(shell jq -r '.wrappers[] | select(.name=="$(WRAPPER)") | .environment // empty' "$(RUN_CONFIG_JSON)" ))
 	@opam update
 	opam install --switch=$* --keep-build-dir --yes rungen orun
 	opam install --switch=$* --best-effort --keep-build-dir --yes $(PACKAGES) || $(CONTINUE_ON_OPAM_INSTALL_ERROR)
 	opam exec --switch $* -- opam list
+	opam list --rec --resolve ocaml-migrate-parsetree
```
shubhamkumar13, line 120: remove this

shubhamkumar13:
@shakthimaan I added a check for `+effects` in Makefile to ensure this works with `4.12+domains` and `4.12+domains+effects`.

run_config.json
```diff
@@ -18,6 +18,30 @@
     }
   ],
   "benchmarks": [
+    {
+      "executable": "benchmarks/irmin/layers.exe -n 16000",
+      "name": "layers_maxrss",
```
kayceesrk, line 23: Is there a significance behind naming this as `layers_maxrss`? Separately, I'd recommend using informative names because these tend to show up on the graphs. `layers_*` is not illustrative. What does layers do? Perhaps you can capture the purpose in a word or two.

run_config.json
```diff
@@ -18,6 +18,30 @@
...
+      ],
+      "runs": [
+        {
+          "params": ""
+        }
+      ]
+    },
+    {
+      "executable": "benchmarks/irmin/layers.exe -b 100 -n 100",
+      "name": "layers_majorgc",
```
kayceesrk, line 35: Is there a significance for the name `layers_majorgc`?

kayceesrk:
Thanks. I have had a look at the PR quickly. Here are my comments:

* The description of the PR is currently empty. It is always a good idea to include a description in PRs.
* Please include the different things the PR does. The PR has lots of changes (41 files changed). I can at least see the addition of layers benchmark, updating the dependent packages (?), ppx support, etc.

kayceesrk (changes requested):

shubhamkumar13:
This PR adds the layers benchmark to the sandmark suite. The layers benchmark is one of the benchmarks present in the [mirage/irmin/bench/irmin-pack](https://github.com/mirage/irmin/blob/master/bench/irmin-pack/layers.ml) repo. 

The benchmark (as far as I can tell) creates a store of type [`Irmin_pack_layered`](https://mirage.github.io/irmin/irmin-pack/Irmin_pack_layered/index.html) and generates a tree structure for that layered store, checks out the branch of the tree and commits to the store. The stats for this complete cycle is calculated (for each commit) and it prints out those stats after the run.

The current PR adds the layers benchmark and it's dependencies to sandmark. The layers benchmark contains a lot of ppx code so the layers.ml benchmark file added to sandmark contains ppx generated code to avoid preprocessing on every sandmark run. 

But we do require ppx libraries and one of the major reasons we had to [modify](https://github.com/shubhamkumar13/sandmark/blob/d89647d82c1449b3b75de93584183f93b3c3b03c/Makefile#L115) the Makefile for ppx support. Ppx libraries require omp and ppxlib so we had to create a conditonal check for different variants of ocaml (multicore/stock/domains/domains+effects) to install `ocaml-migrate-parsetree` and `ppxlib` which compile with that ocaml variant. This work was only possible because @Sudha247 was able to create `ocaml-migrate-parsetree` and `ppxlib` versions which compiled with the multicore variant of ocaml.

Addressing the question of why there are 2 different versions of layers benchmark - This is might not be the right way to do it and I would like to address why I went this way.

When I was testing in what ways I can stress the compiler there were 2 things which we wanted to test - larger maxRSS and major cycles with smaller running times.

So I remember testing this with `-n` flag as an input and increasing the parameters associated with the flag from 1k to 16k. (the -n flag is for assigning the number of commits to the store in the layers benchmark).

And maximum amount of major allocations that the benchmark did for 16k commits was 278 for `4.10.0+multicore` and 295 for `4.10.0+stock`. And the maxRSS value for 16k commits were around 8.6 GB for `4.10.0+multicore`. They could have both been used together in that same layers entry as 16k commits were producing both these results. 

So why did I create another entry? (And also addressing why that was short sighted and wrong when I look back at it)

If we see the executable line for `layers_majorgc` it looks like this -> `   "executable": "benchmarks/irmin/layers.exe -b 100 -n 100",`
It uses a `-n` flag and a `-b` flag which is different from the `layers_maxrss` entry which looks like this -> `      "executable": "benchmarks/irmin/layers.exe -n 16000",`

Sidenote (To address what a `-b` flag is) : This is used to change the number cycles where every cycle proceeds to commit `-n` number of times to the store. 
For Example, in the above statement `-b 100 -n 100` means that there will be 100 cycles, each cycle would have 100 commits to the store. 

I was not sure how the `params` would work with different flag entries so I created another entry instead, which might not have been the best solution.

And there might be another question arising from the above statement, I'm assuming here, "Why I went with a different set of flag configurations?" 
And the answer is to decrease the running time and to increase the number of major allocations. Which might have been wrong now that I look back at my decision. 

So the command `benchmarks/irmin/layers.exe -n 16000` as I mentioned before did 278 (on `4.10.0+multicore`) major collections but took `3523.36s` that's around around an hour. But `benchmarks/irmin/layers.exe -b 100 -n 100` ran for `553.75s` (which is around 9 mins) and did 576 (on `4.10.0+multicore`) major collections. And I thought this might be helpful if we individually wanted to stress the major gc. 

But it might be take up more time if run it sequentially which is the reason why I'm thinking this might be wrong. So if this needs to be changed then I would happily change this :smile:  as it seems to not help in reducing running time when ran together.

Also creating unnecessary confusion with the names on the graphs.

kayceesrk:
Thanks for the comment. Fewer configurations to run is better. If there is a possibility of combining the two runs into one then please try to do it. 

Also please update the description of the PR with a summary of your comment. Currently the description is empty.

kayceesrk:
What is the maxRSS on the 10 minute run? The description is useful, but hard to read. 

Can you draw a table with the two configurations, and list all the different observed values (time, major GC, minor GC, maxRSS, etc)? You only need to pick up these results from the bench file of an earlier run. This would ease the decision making.

shubhamkumar13:
Apologies for this being late, So the following table might help to understand the statement I mentioned above.
 - with 4.10.0+stock
 
| benchmark | time (secs) | maxRSS (kB) | major GC | minor GC |
|----------------|------|---------------|--------------|---------------|
|layers_majorgc| 573.7721099853516 | 647884 | 591 | 214538 |
|layers_maxrss| 3528.1806740760803 | 7531440 | 306 | 1247957 |

 - with 4.10.0+multicore
 
| benchmark | time (secs) | maxRSS (kB) | major GC | minor GC |
|----------------|------|---------------|--------------|---------------|
|layers_majorgc| 592.2564759254456 | 389064 | 558 | 217448 |
|layers_maxrss| 3566.2439448833466 | 4469280 | 294 | 1258860 |

Also I can combine these 2 configurations into one it seems like placing params in `run_config.json` file in the following way works

```
...
   "runs" : [
      { "params":  " -n 16000" },
      { "params":  " -n 100 -b 100" }
    ]
...
```

If it's required or not can be decided based on the table above, I'm not sure if it's needed because of my previous conclusion that this doesn't help with time which was why I created 2 different configurations in the first place and joining them would not help it.

kayceesrk:
Ok. Good to keep both. Perhaps combining them under the same name is good. Can we use the name "irmin_pack_layered" as it is more descriptive?

Issue #207, opened 2021-02-10
graph500seq: Fatal error: exception Sys_error("kronecker.txt: No such file or directory")


shakthimaan:
The graph500seq benchmarks are failing in Sandmark (commit aa122e6dc6ebc004989b1b04e606c1f7c718c87c, January 24, 2021) for `macro_bench` tag with the following error messages:
```
Executing benchmarks with:
  RUN_CONFIG_JSON=run_config_filtered.json
  RUN_BENCH_TARGET=run_orun  (WRAPPER=orun)
  PRE_BENCH_EXEC=
        orun kernel2.12_10.orun.bench [4.10.0+multicore_1] (exit 2)
(cd _build/4.10.0+multicore_1/benchmarks/graph500seq && /home/shakthi/testing/sandmark/_opam/4.10.0+multicore/bin/orun -o ../../kernel2.12_10.orun.bench -- taskset --cpu-list 5 ./kernel2.exe 12 10)
Fatal error: exception Sys_error("kronecker.txt: No such file or directory")

/tmp/orun875078stderr
        orun kernel3.12_10_2.orun.bench [4.10.0+multicore_1] (exit 2)
(cd _build/4.10.0+multicore_1/benchmarks/graph500seq && /home/shakthi/testing/sandmark/_opam/4.10.0+multicore/bin/orun -o ../../kernel3.12_10_2.orun.bench -- taskset --cpu-list 5 ./kernel3.exe 12 10 2)
Fatal error: exception Sys_error("kronecker.txt: No such file or directory")
```
You can reproduce the error using:
```
$ TAG='"macro_bench"' make run_config_filtered.json
$ RUN_CONFIG_JSON=run_config_filtered.json make ocaml-versions/4.10.0+multicore.bench
```
In `benchmarks/graph500seq/dune`, there is:
```
(alias (name buildbench) (deps kronecker.exe kernel1.exe kernel2.exe kernel3.exe))
```
The `deps` are not built in sequential order, and `kronecker.exe` produces `kronecker.txt`, which is a pre-requisite for `kernel2` and `kernel3`.

PR #209, opened 2021-02-18
Use rule target kronecker.txt and remove from macro_bench

shakthimaan:
The graph500seq benchmarks were failing because of `kronecker.txt: No such file or directory` as mentioned at https://github.com/ocaml-bench/sandmark/issues/207. This PR
* Uses a target rule to build kronecker.txt before running `kernel2` and `kernel3`.
* Removes graph500seq benchmarks from `macro_bench` tag.

### DIFF
--- a/benchmarks/graph500seq/dune
+++ b/benchmarks/graph500seq/dune
@@ -1,4 +1,16 @@
 (executables
 	(names kronecker kernel1 kernel2 kernel3))
 
-(alias (name buildbench) (deps kronecker.exe kernel1.exe kernel2.exe kernel3.exe))
+
+(rule
+	(targets kronecker.txt)
+	(deps (:prog kronecker.exe))
+	(action (with-stdout-to %{targets} (run %{prog} 12 10))))
+
+(alias
+	(name test1done)
+	(deps kronecker.txt kronecker.exe kernel1.exe))
+
+(alias
+	(name buildbench)
+	(deps (alias test1done) kernel2.exe kernel3.exe))
--- a/run_config.json
+++ b/run_config.json
@@ -22,7 +22,7 @@
       "executable": "benchmarks/graph500seq/kronecker.exe",
       "name": "kronecker",
       "tags": [
-        "10s_100s", "macro_bench"
+        "10s_100s"
       ],
       "runs": [
 	{
@@ -34,7 +34,7 @@
       "executable": "benchmarks/graph500seq/kernel1.exe",
       "name": "kernel1",
       "tags": [
-        "1s_10s", "macro_bench"
+        "1s_10s"
       ],
       "runs": [
 	{
@@ -46,7 +46,7 @@
       "executable": "benchmarks/graph500seq/kernel2.exe",
       "name": "kernel2",
       "tags": [
-        "10s_100s", "macro_bench"
+        "10s_100s"
       ],
       "runs": [
 	{


kayceesrk:
What's the thought process behind removing the `macro_benchmark` tag? Do these benchmarks not run long enough?

shakthimaan:
> What's the thought process behind removing the `macro_benchmark` tag? Do these benchmarks not run long enough?
Yes, they do not run long enough, and hence the benchmark graphs are getting skewed. Hence, the request to remove them from `macro_bench`.

kayceesrk:
If that's the case then the timing tags `1s_10s` etc should also be removed?

Sudha247:
This is one of the results I got a while ago of graph500 benchmarks:
```
{"name":"kernel1.12_10","command":"taskset --cpu-list 5 ./kernel1.exe 12 10","time_secs":0.0026328563690185547,"user_time_secs":0.002383,"sys_time_secs":0.0,"maxrss_kB":5600,"ocaml":{"version":"4.10.0+multicore","c_compiler":"gcc","architecture":"amd64","word_size":"64","system":"linux","stats":"false","function_sections":"true","supports_shared_libraries":"true"},"gc":{"allocated_words":1874,"minor_words":1808,"promoted_words":0,"major_words":66,"minor_collections":0,"major_collections":0,"heap_words":4096,"top_heap_words":4096,"mean_space_overhead":0.0},"codesize":206811.0,"ocaml_url":"https://github.com/Sudha247/ocaml-multicore/archive/3deeb2604320a7337cc75f20116ea237d65fc400.tar.gz"}

```
Might be the case that the benchmark was not running due to an absent input file and this PR fixes it. But I did not see an error message saying the benchmark failed to run, it's possible the error message was not printed. We may have to revisit the timing tags for this. Also, wrt `macro_bench` tag, it will be good to confirm the results are not fluctuating (which I observed a few times) before adding it again as a macro benchmark.

kayceesrk:
Can we confirm the running times before removing the `macro_bench` tag? If it does run for the amount of time it says it does, then I'd prefer to keep the `macro_bench` tag. 

If there is variance between different runs, it may be due to the random number generation. We should fix the seed to a constant value if that is the case.

Sudha247:
I'm happy with retaining the tag if the benchmarks are running long enough. So, I think what needs to be done is:
* Running the benchmarks with the target rule in this PR and getting the times for graph500 benchmarks, I believe this would give us accurate timing data. Once we have the results, timing tags might need to be adjusted accordingly.
* Running the benchmarks some number of times to be sure there isn't much variation between different runs. All good if there is not much variation, in case there is this could become a source of noise in overall data and we need to find a way to fix it.

kayceesrk:
My assessment is that we should leave the `macro_bench` tag in. kernel1 runs for about 4s (with changes; see below), kernel2 runs for 25 seconds, and kernel3 runs for 80 seconds. They also do a fair number of major and minor collections. 

Looks like `kernel1` does not actually run the main function, but only defines it. See https://github.com/ocaml-bench/sandmark/blob/master/benchmarks/graph500seq/kernel1.ml#L106-L113. It is unsurprising it finishes immediately. 
An easy fix would be to rename `kernel1.ml` to `graph_construction.ml` and implement a new `kernel1.ml` that calls `linkKronecker()` function. Reading the command line arguments for `scale` and `edgefactor` should be moved to the new `kernel1.ml` from the `graph_construction.ml`. This would of course mean that `linkKronecker` function would take scale and edge factors as arguments. 

It seems weird that all the kernels take command-line arguments, currently set as `12 10` in `run_config.json`, and the kernels in addition take the `kronecker.txt`. Currently, `kronecker.txt` is generated from `kronecker.exe 12 10`. The interface seems brittle. Surely, we shouldn't have to pass `12 10` as arguments to each of the executables. What would happen if I generate `kronecker.txt` with arguments `15 15` and then run the kernels with `12 10`. Something must go wrong? If not, why? The whole thing smells a little fishy. 

Kernel2 and kernel3 write to the standard output. We should redirect the output to `/dev/null` in `run_config.json`. 

We should use snake case and not camel case to be consistent.

PR #211, opened 2021-03-04
Update README.md

sadiqj:


README.md
```diff
@@ -87,7 +87,10 @@ The various options are described below:
   specific flags to the `configure` script.
 
 - `runparams` is OPTIONAL, and its values are passed to OCAMLRUNPARAM
-  when building the compiler.
+  when building the *compiler*.
```
kayceesrk, line 90: We should be explicit and say that these parameters would be used while running the programs.

README.md
```diff
@@ -87,7 +87,10 @@ The various options are described below:
...
+  
+- `environment` is OPTIONAL, and it can be used to set environment variables
+  when running *benchmarks*. (e.g ```ENVIRONMENT="LD_PRELOAD=.. OCAMLRUNPARAM='a=2'```)
```
kayceesrk, line 93: If I define `OCAMLRUNPARAM` in environment and also a give `runparams`, what happens? Which gets overridden. It might be useful to document that here with a specific example.

Consider `runparams` to be `a=2,v=0x400` and `environment` to have `OCAMLRUNPARAM='v=0x800'`. What is the actual list of params passed to the benchmark run? Would it be 

a. v=0x800
b. a=2,v=0x400
c. a=2,v=0x800

@shakthimaan

shakthimaan: The `environment` should be moved under the `Configuring the benchmark runs` section. The `runparams` is used when passed when building the compiler, whereas, `environment` is passed when executing the benchmarks. The objective is the same, but, the context is different.

README.md
sadiqj: Are they? As far as I can see `runparams` only applies when building the compiler itself:

https://github.com/ocaml-bench/sandmark/blob/master/Makefile#L93

README.md
sadiqj: I think https://github.com/ocaml-bench/sandmark/blob/master/Makefile#L110 and https://github.com/ocaml-bench/sandmark/blob/master/orun/orun.ml#L253 together mean we'd get a magical fourth option of `a=2,v=0x800,v=0x400`

sadiqj: How much of a pain is it going to be if we rename these?

If I just saw `runparams`, I would be surprised that it only applied to building the compiler.

shakthimaan: > How much of a pain is it going to be if we rename these?

What should these be renamed to? The reason for the present naming convention is from https://github.com/ocaml-bench/sandmark/issues/184#issuecomment-716485989.

sadiqj:
Looking at the PR for allocation benchmarks, this update is actually wrong!

Issue #142, opened 2020-07-16
[RFC] How should a user configure a sandmark run?


ctk21:
If we look at things as a user wanting to run a benchmarking experiment, then they are wanting to compose several components to get data. The components they want to compose are:
- A compiler built with their favourite configure switches
- A set of benchmarks they want to run (expressed preferably as the end binary rather than needing to know the dependencies)
- A program that collects stats from the binary (e.g. perf, orun, binary size)
- An environment configuration for the execution of the stats collection (e.g. taskset, ASLR, OCAMLRUNPARAM)

The user configures (and could do so implicitly from defaults) these components into a run plan of benchmark executions to get data.

The mechanisms for configuring up the above has evolved as we’ve tried to do more with sandmark and (currently) is spread out:
 - the compiler build is specified in the .comp file
 - the benchmarks and their parameters are in run_config.json but you have to get the `BUILD_BENCH_TARGET` correct (e.g. multicore vs serial benchmarks) for all the opam packages because the benchmark programs don’t bring their dependent opam packages with them.
 - the stat collection wrappers are in run_config.json
 - the environment configuration is a bit all over the place. Some of it is going on in the wrappers in run_config.json, we have some stuff happening (e.g. crontab scripts) with PRE_BENCH_EXEC and we have `PARAMWRAPPER` handling multicore tasksetting

With this issue I'm hoping to find out if this tuple of: 
 ```
(
 compiler build, 
 set of benchmarks (incl dependencies),
 executable to collect stats from a benchmark command, 
 environment within which to run the executable
)
```
covers all the use cases we need?

Once we've got the use cases nailed down. We can try to have proposals or prototypes that attempt to make the configuration a bit easier to handle than the current mix of environment variables, comp files and json files. This should make it easier for users to run the experiments they want.

shakthimaan:
Some thoughts:

* It will be good to categorize the benchmarks, so that a user can simply include a family of benchmarks for an experiment or a test. For example, a researcher interested in studying graph algorithms or analyzing performance for time-series data can run only those relevant benchmarks.

* All the configuration should exist in a `config/` folder in the project repository, so that, the user knows as to where they need to look for in order to make a change.

* For any test, it will be good to allow the user to setup one or more `profiles` or `experiments` which describes in detail the relevant parameters used for benchmarking.

* It will be useful to have everything included in a single configuration file, including environment variables, so that, using a given file, we should be able to re-run any experiment.

[...]

PR #212, opened 2021-03-11
Increasing the major heap allocation on some benchmarks

sadiqj:
This PR isn't meant to be merged as-is but should serve as a starting point for adding more longer running benchmarks to Sandmark that involve some major heap work. 

For some benchmarks already in Sandmark I've either increased the parameters or I've added loops around their main loop to make them run longer.

For benchmarks with increased parameters we should split the new parameters out in to a separate run (in run_config) with an appropriate tag for allocator focused benchmarks.

For benchmarks where we've added a loop around, we'll need to parameterise that loop and then make sure the existing runs include that parameter, with a new set of runs with the allocator focused tag.

@shubhamkumar13 as mentioned on Slack.

@shakthimaan this PR also includes the wrapper change (and run_serial shows how I end up running them) which is probably an interesting use case that you might want to incorporate in to your sandmark 2.0 work. It's not the easiest of use cases to make work at the moment.


benchmarks/benchmarksgame/regexredux2.ml
```diff
@@ -53,8 +53,10 @@ let count re s =
 
 
 let () =
+  for a = 0 to 20 do
```
shakthimaan, line 56: 1. Any specific reason to use the number 20? 
2. Is it useful to parameterize the same?

run_all_serial.sh
```diff
@@ -2,17 +2,9 @@
...
-RUN_BENCH_TARGET=run_pausetimes_multicore RUN_CONFIG_JSON=run_config_filtered.json make ocaml-versions/4.06.1+multicore+pausetimes.bench
-
-RUN_CONFIG_JSON=run_config_filtered.json make ocaml-versions/4.06.1+multicore+stw.bench
-RUN_BENCH_TARGET=run_pausetimes_multicore RUN_CONFIG_JSON=run_config_filtered.json make ocaml-versions/4.06.1+multicore+stw+pausetimes.bench
-
-RUN_CONFIG_JSON=run_config_filtered.json make ocaml-versions/4.06.1+stock.bench
-RUN_BENCH_TARGET=run_pausetimes_trunk RUN_CONFIG_JSON=run_config_filtered.json make ocaml-versions/4.06.1+stock+instrumented.bench
+RUN_BENCH_TARGET=run_orun_bf80 RUN_CONFIG_JSON=run_config_filtered.json make ocaml-versions/4.10.0+stock.bench
+RUN_BENCH_TARGET=run_orun_bf100 RUN_CONFIG_JSON=run_config_filtered.json make ocaml-versions/4.10.0+stock.bench
+RUN_BENCH_TARGET=run_orun_bf120 RUN_CONFIG_JSON=run_config_filtered.json make ocaml-versions/4.10.0+stock.bench
```
shakthimaan, line 7: What does `bf` stand for?

run_config.json
```diff
@@ -4,6 +4,27 @@
...
+	      "command": "orun -o %{output} -- taskset --cpu-list 5 %{command}"
+    },	
+    { 
+	      "name": "orun_bf120",
+	      "environment": "OCAMLRUNPARAM='a=2,o=120'",
+	      "command": "orun -o %{output} -- taskset --cpu-list 5 %{command}"
+    },
+    { 
+	      "name": "orun_mimalloc",
+	      "environment": "LD_PRELOAD=/home/sadiq/sandmark/libmimalloc.so.1.7",
```
shakthimaan, line 24: This needs to be updated to use the *nix distribution environment path.

benchmarks/benchmarksgame/regexredux2.ml
sadiqj: 1. No, it was essentially the minimum iterations to get the runtime long enough that it ran for five minutes or so.
2. Yea, I think that's a good idea - we can make even longer running versions of these benchmarks if we need to.

run_all_serial.sh
sadiqj: The Best Fit allocator from trunk.

run_config.json
sadiqj: Interesting. Is this documented anywhere?

shakthimaan: From where is this `libmimalloc` installed? Can you please elaborate?

sadiqj: Oh, libmimalloc is built separately.

The reason for this PR is not that I want mimalloc specifically available in sandmark but to show a sandmark usecase that we've had recently which is a bit of a pain to do.

As you can see right now every combination of environment or OCAMLRUNPARAMs requires a separate wrapper to be defined. It would be nice if there was an easier of way of saying "I want this wrapper with this environment".

shakthimaan: Can you please provide the steps to build `libmimalloc`?
> It would be nice if there was an easier of way of saying "I want this wrapper with this environment".
Will address it when working on the User Configuration feature request at https://github.com/ocaml-bench/sandmark/issues/142.

sadiqj: Sure. The repo is at https://github.com/microsoft/mimalloc . I just followed the Linux build instructions, that should build you the library which you can copy.

PR #220, opened 2021-05-04
Attempt to improve the OCAMLRUNPARAM documentation

ctk21:
This PR attempts to document the `environment` parameter that was introduced in #189 and make things clearer for people running experiments which use `OCAMLRUNPARAM` settings.

It is an update of #211 which didn't end up merging.


shakthimaan:
LGTM.

PR #221, opened 2021-05-05
Fix up decompress iterations of work

ctk21:
This PR fixes errors between the amount of work configured and the amount of work done. 

The changes are:
- fix up for indices and explicitly count the amount of work done with an assert that the work done is the work requested. (consider `iterations=4` & `num_domains=2`, in this configuration I make there being _3_ units of work being done by each of the domains and not 2 in the existing code).
- moves the multicore test to use a pool with `parallel_for`; this is important as there are unintended consequences in the existing code (consider `iterations=3` & `num_domains=4`, it will distribute `work 0` to the first three spawned domains and then do 3 units of work in the final domain).
- simplifies the `data_to_compress` setup to use `String.init` which is more idiomatic.

### DIFF
--- a/benchmarks/decompress/dune
+++ b/benchmarks/decompress/dune
@@ -6,7 +6,7 @@
 (executable
  (name test_decompress_multicore)
   (modules test_decompress_multicore)
-  (libraries bigstringaf checkseum.ocaml decompress.zl))
+  (libraries bigstringaf checkseum.ocaml decompress.zl domainslib))
 
 (alias (name buildbench)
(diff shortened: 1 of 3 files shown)


sadiqj:
This is a good simplification and I don't see any issues with the changes.

I'm interested in how the parallel changes perform against what we have at the moment.

sadiqj (approved):

ctk21:
Here are benchmarks (detuned Zen2 machine) for this change:
20210505_pr221_time
20210505_pr221_speedup

PR #226, opened 2021-05-18
Better scaling for mandelbrot6_multicore

ctk21:
This PR improves the `mandlebrot6_multicore` scaling. It does this by:
 - Using `parallel_for` in the work distribution
 - Minor indent changes to make closer to the single domain version

Results from a (detuned) Zen2:
20210518_mandel_time
20210518_mandel_speedup
20210518_mandel_mgc

You could imagine going further than this by exploiting parallelism between the `output_bytes` and the work completion. However just using the `parallel_for` divide-and-conquer with default `chunk_sizes` is beating the existing code at 4 cores and more by a meaningful margin.


kayceesrk:
👍

PR #239, opened 2021-06-15
Port grammatrix to Task pool

Sudha247:
This patch moves the multicore grammatrix benchmark to Domainslib Task pool.

Results for different chunk sizes of `parallel_for`, 0 being the default chunk size. Old indicates the handwritten version currently in sandmark.
image

### DIFF
--- a/benchmarks/multicore-grammatrix/grammatrix_multicore.ml
+++ b/benchmarks/multicore-grammatrix/grammatrix_multicore.ml
@@ -8,11 +8,11 @@
 
 module A = Array
 module L = List
-module C = Domainslib.Chan
+module T = Domainslib.Task
 
 let num_domains = try int_of_string Sys.argv.(1) with _ -> 4
 let input_fn = try Sys.argv.(3) with _ ->  "data/tox21_nrar_ligands_std_rand_01.csv"
-let chunk_size = try int_of_string Sys.argv.(2) with _ -> 16
+let chunk_size = try int_of_string Sys.argv.(2) with _ -> 0
 
 let dot_product xs ys =
   let n = A.length xs in
@@ -23,16 +23,17 @@ let dot_product xs ys =
   done;
   !res
 
-let compute_gram_matrix samples res s e =
+let compute_gram_matrix samples pool =
   let n = A.length samples in
   assert(n > 0);
-  for i = s to e do
+  let res = A.init n (fun _ -> A.create_float n) in
+  T.parallel_for ~start:0 ~finish:(n - 1) ~body:(fun i ->
     for j = i to n - 1 do
       let x = dot_product samples.(i) samples.(j) in
       res.(i).(j) <- x;
       res.(j).(i) <- x (* symmetric matrix *)
-    done
-  done
+    done) pool;
+  res
 
 let parse_line line =
   let int_strings = Utls.string_split_on_char ' ' line in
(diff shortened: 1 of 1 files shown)


Sudha247:
The task pool is now created after all inputs are received from data files. @ctk21  rightly pointed out that when pool is created at the very beginning, the number of minor GCs are quite high. Now the number of GC cycles is close to the hand written version, however there's still some slowdown as the cores increase.

Updated results:
**Speedup**
image
**Time**
image

ctk21:
There's something interesting going on that I can't explain!

Couple of possibilities:
- GC differences; although it looks like we have resolved those.
- static work distribution with a channel is just better than the dynamic distribution with a workstealing deque; this is possible but would surprise me because the work units look fairly long running (~5s with 24 cores is ~200ms per core over the default 8 units is ~26ms per work unit). 
- workload ordering; the units of work in `parallel_for` look imbalanced (look at the loop over `j` with `i = n-2` vs `i = 0`), maybe somehow we are doing all the short units first and then have the long units at the end which leads to idle domains and a poor critical path?

Maybe forcing `parallel_for` to have `{0..n_domains}` units of work where we recreate exactly the partition in the `Chan` implementation might tell us something?

Sudha247:
Thanks for the comments @ctk21!

>GC differences; although it looks like we have resolved those.

I presume they are resolved. The GC stats for the two versions -
**Old (Channels version)**

```
allocated_words: 213291844
minor_words: 146565640
promoted_words: 1906480
major_words: 68632684
minor_collections: 831
major_collections: 15
forced_major_collections: 0
heap_words: 63564882
top_heap_words: 64420946
mean_space_overhead: 4.709006
```

**New (Task pool version)**

```
allocated_words: 213285937
minor_words: 146559764
promoted_words: 1910484
major_words: 68636657
minor_collections: 709
major_collections: 15
forced_major_collections: 0
heap_words: 63691858
top_heap_words: 64560210
mean_space_overhead: 4.706600
```

>workload ordering; the units of work in parallel_for look imbalanced

Right, so parallel_for would take care of the task distribution to an extent. Despite that some disorderliness might have crept in here. @ctk21's suggestion of clubbing together loop index i and (n - i) to get a roughly equal workload on each domain takes care of maintaining almost equal work distribution. This improves the program a bit on higher number of cores, though it's a little lagging from the chan implementation.

**Time:**

image

**Speedup**

image

ctk21:
Nice to see things closer, looks like the ordering of imbalanced units is important in this example.

I'm not 100% convinced the new code is always doing the same amount of work:
```ocaml
  T.parallel_for ~start:0 ~finish:((n - 1)/2) ~body:(fun i ->
     for j = i to n - 1 do
       let x = dot_product samples.(i) samples.(j) in
       res.(i).(j) <- x;
       res.(j).(i) <- x (* symmetric matrix *)
    done;
    for j = (n - 1 - i) to (n - 1) do
      let x = dot_product samples.(n - 1 - i) samples.(j) in
      res.(n - 1 - i).(j) <- x;
      res.(j).(n - 1 - i) <- x (* symmetric matrix *)
     done) pool;
```
If you take `n = 7`, then I see `j = 3` being executed in both of the `j` loops when `i = (7-1)/2`

Sudha247:
>If you take n = 7, then I see j = 3 being executed in both of the j loops when i = (7-1)/2

That's right, and it turns out the sequential version does the same too. So the amount the number of times `res` is modified is the same in all of the versions.

That said, retiring this in favour of the current version. (ref. #242)

PR #241, opened 2021-06-25
Switch to default Random module

Sudha247:
Default Random module stores its state in Domain-local storage (see https://github.com/ocaml-multicore/ocaml-multicore/pull/582), thereby making it convenient to be called from multiple domains. Retaining `Random.State` in minilight which uses `Random.State` in the sequential version, global roots micro benchmarks and evolutionary algorithm.

### DIFF
--- a/benchmarks/multicore-lockfree/test_bag_multicore.ml
+++ b/benchmarks/multicore-lockfree/test_bag_multicore.ml
@@ -6,19 +6,17 @@ let entries_per_dom = num_entries / num_domains
 module Bag = Lockfree.Bag
 
 let bag = Bag.create ()
-let k = Domain.DLS.new_key Random.State.make_self_init
 
 let add_or_remove_entries n () =
-  let state = Domain.DLS.get k in
   for _ = 1 to n do
-  let r = Random.State.int state 100 in
+  let r = Random.int 100 in
   if (r > read_percent) then
-    Bag.push bag (Random.State.int state n)
-  else 
+    Bag.push bag (Random.int n)
+  else
     Bag.pop bag |> ignore
   done
-  
-let _ = 
+
+let _ =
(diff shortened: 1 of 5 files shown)


ctk21:
I'm not convinced this is a good idea. The problem is that the default `Random` will initialize the sequences to be _identical_ in each domain. The pattern we currently have uses:
```ocaml
let k = Domain.DLS.new_key Random.State.make_self_init
```
which will ensure that each domain will generate a different sequence. 

In terms of benchmarking it might not be a big deal, however if people copy the pattern then they may get upset when they discover that each domain is receiving an identical stream of numbers.

For example, imagine you use the naive `Random` in different domains to intitialize the rows of a matrix or setup sequences for a monte-carlo calculation, something like:
```ocaml
let n = 4 in
let init_fn () = Array.init n (fun _ -> Random.int 1234) in
let ds = Array.init n (fun _ -> Domain.spawn init_fn) in
let output d = Array.iter (Printf.printf "%i ") (Domain.join d); Printf.printf "\n" in
Array.iter output ds
```
This will generate
```
58 335 146 461
58 335 146 461
58 335 146 461
58 335 146 461
```
that might surprise a user who's intent was that each sequence in each domain is "random".

Sudha247:
I played around with this a bit and figured we can randomize the values generated by different domains by initializing it with a distinct seed via calling `Random.self_init`.

```ocaml
let n = 4 in
let init_fn () = Random.self_init (); Array.init n (fun _ -> Random.int 1234) in
let ds = Array.init n (fun _ -> Domain.spawn init_fn) in
let output d = Array.iter (Printf.printf "%i ") (Domain.join d); Printf.printf "\n" in
Array.iter output ds
```

produces:

```
561 1185 1073 645 
357 676 986 292 
6 1212 1138 594 
843 289 1092 987
```

That said, I think this is quite similar to what we currently do, so maybe it's best to leave it as is.

PR #242, opened 2021-06-29
Add commentary on grammatrix

Sudha247:
The grammatrix benchmark does customised task distribution via channels. Replacing it with `parallel_for` didn't quite yield the same results. This patch adds some comments on the possible reasons with some observations from #239.


kayceesrk:
LGTM.

PR #244, opened 2021-07-15
add chrt to pausetimes_multicore wrapper

shubhamkumar13:
The PR adds `chrt` to the `pausetimes_multicore` wrapper to make sure the taskset argument mentioned in the parawrapper is picked up.

The absence of `chrt` led to the benchmarks not pick up the taskset arguments resulting in slow nightly runs.

Thanks to @Engil he was able to pinpoint the problem to the cores not being used.

### DIFF
--- a/multicore_parallel_navajo_run_config.json
+++ b/multicore_parallel_navajo_run_config.json
@@ -14,7 +14,7 @@
     },
     {
       "name": "pausetimes_multicore",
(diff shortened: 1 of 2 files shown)


kayceesrk:
Which is the correct order?

image

kayceesrk:
Best to keep `chrt -r 1` after `paramwrapper`.

PR #230, opened 2021-05-27
[WIP] Build for 4.13.0+trunk with dune.2.8.1

shakthimaan:
The dependency packages have been updated to build 4.13.0+trunk with locally installed dune.2.8.1. This builds at least 56 of the macro benchmarks using:
```
TAG='"macro_bench"' make run_config_filtered.json
RUN_CONFIG_JSON=run_config_filtered.json make ocaml-versions/4.13.0+trunk.bench
```

PR #245, opened 2021-07-15
Add dune.2.9.0

Engil:
After investigating the hanging of the `coq` build with Multicore OCaml, under some specific scenario, it turns out updating `dune`  to 2.9.0 fixes the issue.
This specific hang is still a bit of a mystery to me, but it seems like the failure to build `coq` with Multicore triggers a loop in a retry mechanism in `dune`'s fibers.

I also noticed that we still have a _system dune_ in place to work around past compatibility issues with `4.09.0` and Multicore not being able to build `dune` by itself.

It does seems the dune version used by the sandmark run is indeed the one installed in the switch. (and not the system dune)

I was wondering if there was a reason why we still have the system dune setup active in sandmark? I assume it is still useful for trunk (since dune may break against trunk), however Multicore+dune compatibility is not an issue anymore.

### DIFF
--- a/dependencies/packages/dune/dune.2.9.0/opam
+++ b/dependencies/packages/dune/dune.2.9.0/opam
@@ -1,0 +1,53 @@
+x-commit-hash: "641a95d2254ca7c51c97f07f2eed85b7a95db954"
+opam-version: "2.0"
+synopsis: "Fast, portable, and opinionated build system"
+description: """
+dune is a build system that was designed to simplify the release of
+Jane Street packages. It reads metadata from "dune" files following a
+very simple s-expression syntax.
+
+dune is fast, has very low-overhead, and supports parallel builds on
+all platforms. It has no system dependencies; all you need to build
+dune or packages using dune is OCaml. You don't need make or bash
+as long as the packages themselves don't use bash explicitly.
+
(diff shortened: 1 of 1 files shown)


kayceesrk:
@shakthimaan please can you take a look at this?

Also,

> I also noticed that we still have a system dune in place to work around past compatibility issues with 4.09.0 and Multicore not being able to build dune by itself.

We shouldn't need to hardwire system dune anymore.

shakthimaan:
> Multicore+dune compatibility is not an issue anymore.
Yes. The sys_dune_hack has been removed in the 2.0.0-alpha branch.
Source: https://github.com/shakthimaan/sandmark/blob/2.0.0-alpha%2B001/Makefile

We are still blocked on dune.2.8 as we require the Irmin benchmarks to be built.
https://github.com/ocaml-bench/sandmark/pull/230

kayceesrk:
We should retain the system dune hack since dune is not known to reliable build on trunk. This will break the entire benchmarking setup for trunk if dune fails to build on trunk.

shakthimaan:
Sandmark has been updated to use dune.2.9.0 with the following merge https://github.com/ocaml-bench/sandmark/pull/251, making this PR redundant. Hence, closing the same.

PR #253, opened 2021-09-15
Remove Domain.Sync.poll() from parallel benchmarks 

shakthimaan:
Multicore OCaml programs no longer need Domain.Sync.poll () since safe points have been merged into Multicore OCaml: https://github.com/ocaml-multicore/ocaml-multicore/pull/573. The PR removes the calls to `poll` from the parallel benchmarks in Sandmark.


kayceesrk:
I've updated the PR title and added some context in the description.

PR #254, opened 2021-09-24
Disable sandboxing.

kayceesrk:
It is unclear whether sandboxing helps in sandmark as sandmark uses a
local opam repo distinct from the global one. Moreover, disabling
sandboxing fixes Docker builds and help with nightly runs on navajo.


shakthimaan:
LGTM.

PR #267, opened 2021-12-06
Added support for bench.Dockerfile

shakthimaan:
Added `bench.Dockerfile` and updated Makefile in order to run Sandmark with `current-bench`.


bench.Dockerfile
```diff
@@ -0,0 +1,18 @@
...
+RUN sudo apt-get update
+RUN sudo apt-get -y install libgmp-dev libdw-dev jq jo python3-pip pkg-config m4 autoconf
+
+COPY . .
+
+RUN sudo chown -R opam /app
+RUN eval $(opam env)
+
+RUN opam update
+RUN opam install dune.2.9.0
```
art-w, line 16: The `opam update && opam install dune..` could be done before the `COPY . .` so that docker can cache it: the build will go faster! (it they are done after the `COPY`, docker will believe that this commands can produce a different result because the copied files are different)

Makefile
```diff
@@ -215,10 +218,7 @@ json:
 	};
 
 prep_bench:
-	@{ 	TAG='"run_in_ci"' $(MAKE) multicore_parallel_run_config_filtered.json; \
-		TAG='"macro_bench"' $(MAKE) multicore_parallel_run_config_filtered_filtered.json; \
-		$(MAKE) multicore_parallel_run_config_filtered_filtered_2domains.json; \
-		BUILD_BENCH_TARGET=multibench_parallel RUN_CONFIG_JSON=multicore_parallel_run_config_filtered_filtered_2domains.json $(MAKE) ocaml-versions/4.10.0+multicore.bench; \
+	@{ 	$(BENCH_COMMAND); \
 		$(MAKE) json; \
 	} > /dev/null 2>&1;
```
art-w, line 223: The latest version of current-bench can extract the json from stdin, it should not be necessary to `> /dev/null` the stderr (this can help debugging when looking at the logs)

shubhamkumar13 (approved):
Thanks @shakthimaan  for the PR and @art-w for the review. 
Local testing of the dockerfile seems to work without any issues, I think after updating the changes requested by @art-w it's good to merge

bench.Dockerfile
shakthimaan: Moved!

Makefile
shakthimaan: Removed use of /dev/null.

shubhamkumar13:
LGTM!

PR #269, opened 2022-01-02
Added GitHub Action CI for 5.00.0+trunk

shakthimaan:
Added a new `.github/workflows/build.yml` file to test the `5.00.0+trunk` build with GitHub Actions for the -main branch.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -22,10 +22,13 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


shakthimaan:
* GitHub Actions does not recommend to run `sudo` inside Docker containers as part of the workflow for security reasons.
* Also, there are lot of hacks to get containers to run with GitHub Actions. An open issue for even git checkout exists at https://github.com/actions/checkout/issues/211.
Hence, we will continue to use .drone.yml in CI for now.

PR #276, opened 2022-01-27
Re-add 5.00.0+trunk with CI failure:ignore option

shakthimaan:
Adding `5.00+trunk` with the Drone CI `failure:ignore` option so that we will know if trunk fails to build for a PR, but it should not prevent a PR in getting merged if `5.00+stable` builds fine.

### DIFF
--- a/ocaml-versions/5.00.0+trunk.json
+++ b/ocaml-versions/5.00.0+trunk.json
@@ -1,14 +1,16 @@
 {
   "url" : "https://github.com/ocaml/ocaml/archive/trunk.tar.gz",
   "package_overrides": [
-    "camlpdf.2.3",
-    "cpdf.2.3",
-    "fmt.0.9.0"
+    "fmt.0.9.0",
+    "ocamlfind.1.9.3"
   ],
   "package_remove": [
+    "camlpdf",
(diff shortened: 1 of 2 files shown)


ocaml-versions/5.00.0+trunk.json
```diff
@@ -1,14 +1,16 @@
...
+    "fmt.0.9.0",
+    "ocamlfind.1.9.3"
   ],
   "package_remove": [
+    "camlpdf",
+    "cpdf",
     "coq",
     "coq-core",
     "coq-stdlib",
+    "ctypes",
```
shubhamkumar13, line 13: ctypes works with 5.00.0+trunk after the latest release of bigarray-compat.
Referring - https://github.com/mirage/bigarray-compat/pull/1#issuecomment-1021744418 and the subsequent bumping of the version https://github.com/mirage/bigarray-compat/commit/2ea842ba4ab2cfee7b711f7ad927917f3179a6f9.

I also tested it and it was working with trunk so this might not be needed

shakthimaan: I will create a separate PR for including the latest ctypes.

Issue #274, opened 2022-01-20
Custom Variant Support


shakthimaan:
The compiler developers would like to keep track of any regressions for their development and testing branches, and hence we need a configuration file to store the same.
1. The following information can be stored in a `JSON` file, and should be executed with `sandmark-nightly`.
```
[{
		"github_user": "ocaml",
		"github_repo": "ocaml",
		"branch": "trunk",
		"expiry": "YYYY-MM-DD"
	},
	{
		"github_user": "foo",
		"github_repo": "ocaml",
		"branch": "testing",
		"expiry": "YYYY-MM-DD"
	}
]
```
2. The `expiry` date is the date until which the branch will be tracked and the benchmarks will be executed for the same.
3. A `linter script` should be written to validate the JSON file for any errors in the configuration before it is processed by the CB pipeline.

kayceesrk:
There is an impedance mismatch between the proposed format and the configuration of the compiler build that is used by sandmark: https://github.com/ocaml-bench/sandmark#configuration-of-the-compiler-build. Is there a reason for this divergence? Do you not want it to just be 

```json
[ { "url" : "https://github.com/ocaml-multicore/ocaml-multicore/archive/parallel_minor_gc.tar.gz",
    "configure" : "-q",
    "runparams" : "v=0x400",
    "name": "5.00+trunk+kc+pr23423",
    "expiry": "YYYY-MM-DD"},
  ...]
```

where

* `name`, `url` and `expiry` are mandatory. 
* `name` is what the variant will be displayed as. For example, see how sandmark nightly displays variants as `<name>_<date>_<hash>`.

![image](https://user-images.githubusercontent.com/410484/150334196-e1b122d2-c59c-44ab-8581-eb1580e638a3.png)

The name comes from the variant file. Since we don't have a variant file here, it should be taken explicitly. Essentially a name "5.00+foobar" is the equivalent of having a file named `sandmark/ocaml-versions/5.00+foobar.json`. 

For the sake of completion, I should add that the json file will be hosted in a private Github repo to which access will be provided for the developers.

kayceesrk:
I noticed some additional options now supported in the variant file https://github.com/ocaml-bench/sandmark/pull/271/files#diff-0f01c254e3d4f78f3c955b738cddb57ea49b16e2a46f5b4f940d6f34168de576R3-R7. We should support all of the supported options + `name` & `expiry`.

shakthimaan:
[...]

PR #283, opened 2022-02-11
Add Custom Variant Support feature

shakthimaan:
The PR is for the Custom Variant Support feature request at https://github.com/ocaml-bench/sandmark/issues/274.
* The following parameters (`url`, `tag`, `config_json`, `name`, `expiry`, `environment`, `configure`, `runparams`) are used in an `ocaml-versions/custom.json` file.
* `SANDMARK_NIGHTLY_DIR` can be passed as an environment variable to the `run_all_custom.sh` script.
* The following three URLs are supported:
1. `ocaml/ocaml`
    https://github.com/ocaml/ocaml/archive/trunk.tar.gz
2. A specific commit
    https://github.com/ocaml/ocaml/archive/b73cbbea4bc40ffd26a459d594a39b99cec4273d.zip
3. A developer branch
    https://github.com/sadiqj/ocaml/archive/refs/heads/eventring-pr.zip
* You can test the same using:
```
$ bash run_all_custom.sh
```

### DIFF
--- a/ocaml-versions/custom.json
+++ b/ocaml-versions/custom.json
@@ -1,0 +1,23 @@
+[
+  {
+    "url": "https://github.com/ocaml/ocaml/archive/trunk.tar.gz",
+    "tag": "run_in_ci",
+    "config_json": "run_config_filtered.json",
+    "name": "5.00+trunk+kc+pr23423",
+    "expiry": "2022-02-15"
+  },
+  {
+    "url": "https://github.com/ocaml/ocaml/archive/b73cbbea4bc40ffd26a459d594a39b99cec4273d.zip",
+    "tag": "run_in_ci",
+    "config_json": "run_config_filtered.json",
+    "name": "5.00+trunk+kc+pr23423",
+    "expiry": "2022-02-15"
+  },
+  {
+    "url": "https://github.com/sadiqj/ocaml/archive/refs/heads/eventring-pr.zip",
+    "tag": "run_in_ci",
+    "config_json": "run_config_filtered.json",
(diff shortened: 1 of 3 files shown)


ocaml-versions/custom.json
```diff
@@ -0,0 +1,23 @@
+[
+  {
+    "url": "https://github.com/ocaml/ocaml/archive/trunk.tar.gz",
+    "tag": "run_in_ci",
+    "config_json": "run_config_filtered.json",
+    "name": "5.00+trunk+kc+pr23423",
```
kayceesrk, line 6: The names are all the same in the 3 custom variants. This should be caught by the linter? If not, you will have an issue with storing the results, wouldn't you? I guess you may not be using the names for anything now?

kayceesrk:
Looks reasonable, modulo the comment. Let's move fast on this one, and get to an end-to-end solution. By end to end, I mean the stage where

1. developers submit a variant to a separate Github repo 
2. sandmark nightly picks that up and runs it on the available machines 
3. the developers are able to see the results on https://sandmark.ocamllabs.io

Until then feel free to aggressively merge.

ocaml-versions/custom.json
shakthimaan: Yes, we are not using the names for now.

kayceesrk: > Yes, we are not using the names for now.

FWIW it may be useful to use different names even if we don't use them. We don't want to break a constraint just because we're not checking whether the constraint holds.

PR #292, opened 2022-02-22
Use Task run in parallel benchmarks

shakthimaan:
* Update to use `domainslib.0.4.1`.
* Wrap calls to `parallel_for` and `await` with Task `run`.


kayceesrk:
I had mentioned elsewhere that it would be incorrect to use multiple `run` functions in the same file. But this is wrong. I believe that this is ok, and shouldn't be a problem. With that, the proposed changes look correct to me. CC @Sudha247.

Sudha247:
Yes, this looks good to merge.

PR #296, opened 2022-02-24
Use a single `Task.run` in parallel benchmarks

### LINKED ISSUE #294: LU_decomposition_multicore.exe 4 2048 loops forever
The `LU_decomposition_multicore.exe 4 2048` runs on a single CPU core instead of four CPU cores and loops forever. This occurs after this merge commit https://github.com/ocaml-bench/sandmark/commit/f3fc54cc9d58b53c5a9adabc1cc6944738a3876d, and the process gets killed. Why so?
LU-decomposition-multicore-4-2048-single-core

Sudha247:
Running `Task.pool` inside the iter function was blocking indefinitely, fixes #294.

### DIFF
--- a/benchmarks/multicore-numerical/LU_decomposition_multicore.ml
+++ b/benchmarks/multicore-numerical/LU_decomposition_multicore.ml
@@ -14,9 +14,8 @@ module SquareMatrix = struct
     fa
   let parallel_create pool f : float array =
     let fa = Array.create_float (mat_size * mat_size) in
-    T.run pool (fun _ ->
       T.parallel_for ~start:0 ~finish:( mat_size * mat_size - 1)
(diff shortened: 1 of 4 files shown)


kayceesrk:
Why can't we have a single `T.run` that covers, and remove everything else:

https://github.com/ocaml-bench/sandmark/blob/14023ecbe1886a98bd2525b8ce24702a9abb9219/benchmarks/multicore-numerical/LU_decomposition_multicore.ml#L61-L65

Sudha247:
Yes, I think that'd simplify reading the programs. All parallel benchmarks now have a single `Task.run` call.

shakthimaan:
I tested the PR changes and the `lu_decomposition_multicore` benchmark ran fine. Thanks!

PR #298, opened 2022-02-28
Support to build and run benchmarks in byte mode

shakthimaan:
* Adds `run_config_byte.json` to build and run benchmarks in byte mode.
* The benchmarks `dune` files have been updated to add the modes for native and byte, but, we need to use the `run_config_byte.json` file as follows:
```
$ USE_SYS_DUNE_HACK=1 SANDMARK_CUSTOM_NAME=5.00.0 BUILD_BENCH_TARGET=bytebench RUN_CONFIG_JSON=run_config_byte.json make ocaml-versions/5.00.0+stable.bench 
```
* The `5.00.0+stable.json` file has been updated to February 24, 2022 commit.
* The CI has not been updated for the run_config_byte.json file as the execution is very slow compared to native runs.


kayceesrk:
Should README also be updated with the information for how to run the bytecode benchmarks?

PR #317, opened 2022-03-31
Add checks for loadavg

### LINKED ISSUE #272: Delay benchmark runs if the machine is active
If the machine running the benchmarks is actively being used, there's no point running the benchmarks as all the results would be sabotaged. Consider implementing something similar to [this](https://github.com/kayceesrk/ocamlbench-scripts/blob/master/run-bench.sh#L211-L227) to delay benchmark start.

moazzammoriani:
An attempt to fix #272. @Sudha247 do you think this is a step in the right direction? If yes then I'm not quite sure which JSON file to add a wait time to [here](https://github.com/ocaml-bench/sandmark/commit/f3fe4eb83e331a943c1710c292d69729f73336cd#diff-96191bf6368f1c1192377fb8300a500d299cedcf01e3b1f7005b51faa6c677a6R99). Do you mind helping me out there?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -223,7 +223,6 @@ ocaml-versions/%.bench: depend override_packages/% log_sandmark_hash ocaml-versi
 	   done } > ocaml-versions/.workspace.$(CONFIG_SWITCH_NAME)
 	opam exec --switch $(CONFIG_SWITCH_NAME) -- rungen _build/$(CONFIG_SWITCH_NAME)_1 $(RUN_CONFIG_JSON) > runs_dune.inc
 	opam exec --switch $(CONFIG_SWITCH_NAME) -- dune build --profile=release --workspace=ocaml-versions/.workspace.$(CONFIG_SWITCH_NAME) @$(BUILD_BENCH_TARGET);
-	@./loadavg.sh $(OPT_WAIT) $(START_TIME)
 	@{ if [ "$(BUILD_ONLY)" -eq 0 ]; then												\
 	        echo "Executing benchmarks with:"; 											\
 	        echo "  RUN_CONFIG_JSON=${RUN_CONFIG_JSON}"; 										\
@@ -315,10 +314,11 @@ check_url: check_jq
 	    done;                                        				\
 	};
 
-set_START_TIME:
+load_check:
 	$(eval START_TIME = $(shell date +%s))
+	@./loadavg.sh $(OPT_WAIT) $(START_TIME)
 
-depend: check_url set_START_TIME
+depend: check_url load_check
 	$(foreach d, $(DEPENDENCIES),      $(call check_dependency, $(d), dpkg -l,   Install on Ubuntu using apt.))
 	$(foreach d, $(PIP_DEPENDENCIES),  $(call check_dependency, $(d), pip3 list --format=columns, Install using pip3 install.))
 
--- a/Makefile
+++ b/Makefile
@@ -216,6 +216,7 @@ ocaml-versions/%.bench: depend override_packages/% log_sandmark_hash ocaml-versi
 	   done } > ocaml-versions/.workspace.$(CONFIG_SWITCH_NAME)
 	opam exec --switch $(CONFIG_SWITCH_NAME) -- rungen _build/$(CONFIG_SWITCH_NAME)_1 $(RUN_CONFIG_JSON) > runs_dune.inc
 	opam exec --switch $(CONFIG_SWITCH_NAME) -- dune build --profile=release --workspace=ocaml-versions/.workspace.$(CONFIG_SWITCH_NAME) @$(BUILD_BENCH_TARGET);
+	@./loadavg.sh $(OPT_WAIT) $(START_TIME)
 	@{ if [ "$(BUILD_ONLY)" -eq 0 ]; then												\
 	        echo "Executing benchmarks with:"; 											\
 	        echo "  RUN_CONFIG_JSON=${RUN_CONFIG_JSON}"; 										\
--- a/Makefile
+++ b/Makefile
@@ -223,7 +223,6 @@ ocaml-versions/%.bench: depend override_packages/% log_sandmark_hash ocaml-versi
 	   done } > ocaml-versions/.workspace.$(CONFIG_SWITCH_NAME)
 	opam exec --switch $(CONFIG_SWITCH_NAME) -- rungen _build/$(CONFIG_SWITCH_NAME)_1 $(RUN_CONFIG_JSON) > runs_dune.inc
 	opam exec --switch $(CONFIG_SWITCH_NAME) -- dune build --profile=release --workspace=ocaml-versions/.workspace.$(CONFIG_SWITCH_NAME) @$(BUILD_BENCH_TARGET);
-	$(eval START_TIME = $(shell date +%s))
 	@./loadavg.sh $(OPT_WAIT) $(START_TIME)
 	@{ if [ "$(BUILD_ONLY)" -eq 0 ]; then												\
 	        echo "Executing benchmarks with:"; 											\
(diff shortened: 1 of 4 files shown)


Sudha247:
I believe `run_all_custom.sh` is written for `bash`, so may not work with `sh`. Can you please try `bash run_all_custom.sh`? Here's one place where it's used: https://github.com/ocaml-bench/sandmark-nightly/blob/main/scripts/nightly_turing.sh#L34

moazzammoriani:
I'm still stuck unfortunately. Perhaps I should have started with the following question to begin with.  What do I need to do in order to get `run_alll_custom.sh` to produce the same results as `run_all_serial.sh`?

Sudha247:
All right, maybe you can test your code snippet on `run_all_serial.sh` first? Also, are you seeing the same error even after 

If you want to test the `run_all_parallel.sh` script, please note that the `cpu-list` is hard-coded for our server machines. You might have to modify it to suit your machine. (cf. https://github.com/ocaml-bench/sandmark/blob/main/multicore_parallel_run_config.json#L43)

moazzammoriani:
@Sudha247 I've tried cooking up this new solution. My guess was that the best point in time to apply the load check is right after the dependencies are installed. After perusing through the Makefile I found the right point to insert the load check. However, the problem I have currently encountered is that loadavg remains the same after the initial check within the control flow of the shell script in the Makefile. That is, once the dependencies are installed and it's time to run the part of the script that I have inserted in the Makefile, `$(shell awk '{print 100*$$1}' /proc/loadavg)` keeps returning the same exact loadavg.

moazzammoriani:
I should point out that I was able to try what you had suggested but I felt like it wasn't solving the actual issue. The loadavg increases typically after the dependencies are installed and checking it before they are installed doesn't solve the problem.

moazzammoriani:
I think I cracked it! Feel free to provide any sort of feedback.

run_all_parallel.sh
```diff
@@ -8,8 +8,17 @@
...
+    OPT_WAIT=0
+fi
+
+
 USE_SYS_DUNE_HACK=1 \
                  RUN_BENCH_TARGET=run_orunchrt \
                  BUILD_BENCH_TARGET=multibench_parallel \
                  RUN_CONFIG_JSON=multicore_parallel_run_config_filtered.json \
-                 make ocaml-versions/5.0.0+stable.bench
+                 make OPT_WAIT=$OPT_WAIT START_TIME=$START_TIME ocaml-versions/5.0.0+stable.bench
```
shakthimaan, line 24: This will in turn invoke the `Makefile`. There is no need to update this script.

run_all_serial.sh
```diff
@@ -2,4 +2,12 @@
...
-USE_SYS_DUNE_HACK=1 RUN_CONFIG_JSON=run_config_filtered.json make ocaml-versions/5.0.0+stable.bench
+START_TIME=$(date +%s)
+
+if [ "$1" = "--wait" ]; then
+    OPT_WAIT=1
+else 
+    OPT_WAIT=0
+fi
+
+USE_SYS_DUNE_HACK=1 RUN_CONFIG_JSON=run_config_filtered.json make OPT_WAIT=$OPT_WAIT START_TIME=$START_TIME ocaml-versions/5.0.0+stable.bench
```
shakthimaan, line 13: This will in turn invoke the `Makefile`. Just adding the check in the Makefile is sufficient.

loadavg.sh
```diff
@@ -0,0 +1,23 @@
...
+
+loadavg () {
+    awk '{print 100*$1}' /proc/loadavg 
+}
+
+if [ $OPT_WAIT = 1 ]; then 
+    sleep 60
+    while [ $(loadavg) -gt 60 ]; do 
+        if [ $(expr $(date +%s) - $START_TIME) -gt $(expr 3600 \* 12) ]; then 
+            echo "COULD NOT START FOR THE PAST 12 HOURS; ABORTING RUN" >&2;
```
shakthimaan, line 14: Please avoid all capital letters.

shakthimaan:
The minimal steps to check that the Sandmark build is fine is as follows:
```
opam install dune.2.9.0
eval $(opam env)
make clean
TAG='"run_in_ci"' make run_config_filtered.json
USE_SYS_DUNE_HACK=1 RUN_CONFIG_JSON=run_config_filtered.json make ocaml-versions/5.00.0+trunk.bench
```

moazzammoriani:
Thank you for the feedback @shakthimaan! I have implemented what you suggested.

Makefile
```diff
@@ -216,6 +223,8 @@ ocaml-versions/%.bench: depend override_packages/% log_sandmark_hash ocaml-versi
 	   done } > ocaml-versions/.workspace.$(CONFIG_SWITCH_NAME)
 	opam exec --switch $(CONFIG_SWITCH_NAME) -- rungen _build/$(CONFIG_SWITCH_NAME)_1 $(RUN_CONFIG_JSON) > runs_dune.inc
 	opam exec --switch $(CONFIG_SWITCH_NAME) -- dune build --profile=release --workspace=ocaml-versions/.workspace.$(CONFIG_SWITCH_NAME) @$(BUILD_BENCH_TARGET);
+	$(eval START_TIME = $(shell date +%s))
```
shakthimaan, line 226: These steps can be moved to a separate Makefile target, and added as a pre-requisite in `depend`, even before the start of the build process. If the server is busy, we should not even proceed with the initial setup.

moazzammoriani: If I set the value of `START_TIME` in a dependency in `depend` then wouldn't that set `START_TIME` to be some time before the dependencies are actually installed?

shakthimaan: >  ... some time before the dependencies are actually installed?

Yes. If the server is busy in the first place, we should not even start anything.

Makefile
```diff
@@ -316,7 +315,10 @@ check_url: check_jq
 	    done;                                        				\
 	};
 
-depend: check_url
+set_START_TIME:
+	$(eval START_TIME = $(shell date +%s))
```
moazzammoriani, line 319: @shakthimaan I hope this was what you meant

shakthimaan: You have to move the invocation of the `loadavg.sh` script as well. Together, they constitute the load average check.

moazzammoriani: Ah I see what you mean. I'll admit though that it's not completely clear to me why you'd want to run `loadavg.sh` before the dependencies are installed. 

My own reasoning for placing the invocation to `loadavg.sh` where it was initially is as follows. After the dependencies are installed, loadavg of the machine is surely to rise. The original issue's concern was that the loadavg should be low before the benchmarks are run. Perhaps you're considering something I am not.

shakthimaan: The `loadavg.sh` script has an `exit 10`. If the dependencies are installed and the script exits, there is no proper cleanup. 

> After the dependencies are installed, loadavg of the machine is surely to rise. 

Sure, that is fine. It is the benchmarking job and we expect it to utilize the CPU cores. We just do not want it to start if there is already heavy tasks that are running in the system.

Makefile
```diff
@@ -307,7 +314,11 @@ check_url: check_jq
 	    done;                                        				\
 	};
 
-depend: check_url
+load_check:
+	$(eval START_TIME = $(shell date +%s))
+	@./loadavg.sh $(OPT_WAIT) $(START_TIME)
```
moazzammoriani, line 319: @Sudha247 I have updated the target.

Makefile
shakthimaan: > I'm assuming this is a part of the reasoning that I was asking about

Yes.

Sudha247:
Thanks for your contribution @moazzammoriani! :tada:

PR #328, opened 2022-04-08
Introduce a baseline variable and a check for baseline existence

kayceesrk:
The PR introduces a separate variable called "baseline" for the baseline of the normalised graphs. Earlier, this was hard coded in each of the graphs. The PR also includes a check for whether the variant exists.

image

This is a simple quality of life improvement. We may have done this with interactive feature such as drop-down boxes. But this seemed the easiest fix that improves the workflow of the compiler dev.


shakthimaan:
LGTM.

PR #333, opened 2022-04-12
Add note about --cpu-list in README

### LINKED ISSUE #332: Add note about `--cpu-list` parameter in README
I have noticed that new users often encounter the problem of their Sandmark build failing because of the discrepancy between their machine's number of cpu cores and the hard-coded value of `--cpu-list` in the config files. So I think this is worth pointing out in the README.

moazzammoriani:
Fixes #332.

### DIFF
--- a/README.md
+++ b/README.md
@@ -216,10 +216,13 @@ The scripts for latencies are present in the `pausetimes/` directory. The
 you are using for your current build based on the machine you happen to be
 running the benchmarks on. For example, if your build is to use `run_config.json`
 and the commands contain a parameter such as `--cpu-list 5` then it is being
-assumed that your machine has at least 5 CPU cores. You may want to adjust the
-`--cpu-list` parameter based on whatever number of CPU cores your machine has
-or if you don't know how many cores it has then consider setting the parameter
-to `--cpu-list 1` to use only one CPU core.
+assumed that your machine has a fifth core to use for the build. If your build
+fails then a possible reason might be that your machine doesn't have a fifth
+CPU core. In such a situation you might want to modify the parameter to use
+any number of cores less than or equal to the number of cores your machine
+has. If you do not know the number of cores your machine has then your safest
+option is to modify to the parameter to `--cpu-list 1` to simply use the first
+core.
 
 ### Results
 
--- a/README.md
+++ b/README.md
@@ -212,6 +212,15 @@ The scripts for latencies are present in the `pausetimes/` directory. The
 `pausetimes_trunk` Bash script obtains the latencies for stock OCaml and the
 `pausetimes_multicore` Bash script for Multicore OCaml.
 
+**NOTE**: You may need to modify the `--cpu-list` parameter in the config file
+you are using for your current build based on the machine you happen to be
+running the benchmarks on. For example, if your build is to use `run_config.json`
+and the commands contain a parameter such as `--cpu-list 5` then it is being
+assumed that your machine has at least 5 CPU cores. You may want to adjust the
+`--cpu-list` parameter based on whatever number of CPU cores your machine has
+or if you don't know how many cores it has then consider setting the parameter
+to `--cpu-list 1` to use only one CPU core.
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -212,6 +212,15 @@ The scripts for latencies are present in the `pausetimes/` directory. The
...
 `pausetimes_multicore` Bash script for Multicore OCaml.
 
+**NOTE**: You may need to modify the `--cpu-list` parameter in the config file
+you are using for your current build based on the machine you happen to be
+running the benchmarks on. For example, if your build is to use `run_config.json`
+and the commands contain a parameter such as `--cpu-list 5` then it is being
+assumed that your machine has at least 5 CPU cores. You may want to adjust the
+`--cpu-list` parameter based on whatever number of CPU cores your machine has
+or if you don't know how many cores it has then consider setting the parameter
+to `--cpu-list 1` to use only one CPU core.
```
Sudha247, line 222: `--cpu-list 5` actually runs the benchmark on only one core, which is core number 5. Making it `--cpu-list 1` would run it on core number 1 instead. This line sort of makes it sound like the former case runs the benchmarks on more than one core, which it doesn't. Would be nice to clarify that here.

Sudha247 (commented):
Thanks! This is indeed useful to mention. Just a minor correction suggested below.

README.md
```diff
@@ -212,6 +212,18 @@ The scripts for latencies are present in the `pausetimes/` directory. The
 `pausetimes_trunk` Bash script obtains the latencies for stock OCaml and the
 `pausetimes_multicore` Bash script for Multicore OCaml.
 
+**NOTE**: You may need to modify the `--cpu-list` parameter in the config file
```
shakthimaan, line 215: It is useful to add a note about `--cpu-list` to the README, but, this is not sufficient. One needs to also disable hyperthreading, setup `isolcpus`, remove RCU callbacks etc. as mentioned in the `Notes on hardware and OS settings for Linux benchmarking` section at https://github.com/ocaml-bench/ocaml_bench_scripts.

README.md
```diff
@@ -212,6 +212,18 @@ The scripts for latencies are present in the `pausetimes/` directory. The
...
+you are using for your current build based on the machine you happen to be
+running the benchmarks on. For example, if your build is to use `run_config.json`
+and the commands contain a parameter such as `--cpu-list 5` then it is being
+assumed that your machine has a fifth core to use for the build. If your build
+fails then a possible reason might be that your machine doesn't have a fifth
+CPU core. In such a situation you might want to modify the parameter to use
+any number of cores less than or equal to the number of cores your machine
```
shakthimaan, line 222: One must use only the isolated CPUs.

README.md
```diff
@@ -212,6 +212,18 @@ The scripts for latencies are present in the `pausetimes/` directory. The
...
+has. If you do not know the number of cores your machine has then your safest
+option is to modify to the parameter to `--cpu-list 1` to simply use the first
+core.
```
shakthimaan, line 225: One needs to know the machine on which the cores are running. Usually, the first few cores are non-isolated CPUs, and are used by the OS.

moazzammoriani: This reasoning stems from the fact that it was usually recommended to set `--cpu-list` to `1` by you or @Sudha247. Inference filled the rest of the gaps I suppose. Would you say that I should remove this suggestion of setting it to `1` altogether and instead suggest to set `--cpu-list` to be some non-isolated CPU? Would such a suggestion break too many abstraction barriers for most users?

shakthimaan: Just setting the `--cpu-list` is not sufficient to run the parallel benchmarks. A detailed information on the setup is required, and will be added as part of HACKING.md.

moazzammoriani: Then perhaps this note should either point to HACKING.md or should be removed altogether then?

shakthimaan: A note on `--cpu-list` will be added to HACKING.md.

PR #336, opened 2022-04-14
Add Parallel benchmark mergesort

ernestmusong:


ernestmusong:
Hi @Sudha247 please review this PR for errors and possible fix. Thanks

Sudha247:
I'm on my phone, I'll be able to give a better review when I'm back at my computer. But from what I see in the CI failure, it looks like the Common module is missing. It needs to be imported from the original repo. An easy way to do that would be copying the file to the directory you are currently working on.

ernestmusong:
Hello @sudha after a series of build failures in PR #336 I see that "CCArray_slice module is used in mergesort.ml and common.ml which are inter-dependent. However I found out that "Array_slice" module has been removed as referenced here
 https://github.com/c-cube/ocaml-containers
 The Docs says "There is no replacement at the moment. Please tell us if you need this to be turned into a sub-library."
I assume that CCArray_Slice and Array_Slice are the same and if thus what is the way forward to fix the issue?

Sudha247:
Thanks for the PR @ernestmusong! Good catch! It appears `CCArray_slice` was removed from the `containers` library, but it's still used in the original benchmark.

A quick way to test would be to use an older version of `containers` to validate the benchmark run. Long term solution I think is to rewrite the benchmark without using `CCArray_slice` module.

ernestmusong:
> Thanks for the PR @ernestmusong! Good catch! It appears `CCArray_slice` was removed from the `containers` library, but it's still used in the original benchmark.
> 
> A quick way to test would be to use an older version of `containers` to validate the benchmark run. Long term solution I think is to rewrite the benchmark without using `CCArray_slice` module.

The version of containers used in the original repo is 2.8.1. looking at the dependencies of this container it shows it can not be tested with ocaml version>=4.11 whereas the sandmark repo uses ocaml version 5.0. see below

depends: [
  "ocaml" { >= "4.03.0" }
  "ocaml" { with-test & < "4.11" }
  "dune" { >= "1.1" }
  "dune-configurator"
  "seq"
  "qtest" { with-test }
  "qcheck" { with-test & < "0.14" }
  "ounit" { with-test }
  "iter" { with-test }
  "gen" { with-test }
  "uutf" { with-test }
  "mdx" { with-test & >= "1.5.0" & < "2.0.0" }
  "odoc" { with-doc }
]

ernestmusong:
Hello @Sudha247 after a deep comparison between CCArray and CCArray_slice as used in mergesort and qsort, there exist common methods such as get, set, and lenght as used in the benchmarks. However, qsort seems easier to fix because it contains  just "CCArray_slice.full" that is not present in CCArray. Further research could probably find a replacement for CCArray_slice.full. to make this benchmark running.
See here https://github.com/ckoparkar/ocaml-benchmarks/blob/master/qsort.ml

On the other hand, mergesort contains additional methods  which have no equivalence in CCArray such as CCArray_slice.sub.

Please I would love to add qsort instead of mergesort as soon as I get a replacement for CCArray_slice.full

Issue #331, opened 2022-04-12
Add owl benchmarks


Sudha247:
[Owl](https://github.com/owlbarn/owl) is a scientific computing library developed in OCaml. It'd be nice to add some owl benchmarks to sandmark. This can be broken down into following steps.

* Add owl dependencies and a simple benchmark.
* Add a set representative sequential owl benchmarks
* Performance profile those benchmarks to identify compute-intensive parts.
* Potentially parallelise some benchmarks.

Each of these can be a separate PR/issue.

Some dependencies need to be pinned to make them work with `5.0.0+trunk` compiler variant:
- `owl`: https://github.com/Sudha247/owl/tree/500
- `ocaml-plplot`: https://github.com/Sudha247/ocaml-plplot/tree/500

moazzammoriani:
Hi may I begin working on this?

Sudha247:
Sure, you can work on this.

PR #341, opened 2022-04-21
Add owl 

moazzammoriani:
This is a partly concerning #331.


moazzammoriani:
After much hacking with the `Makefile` variable called `PACKAGES` and also adjusting the pin of the `base` package from `v0.14.3` to `v0.15.0`, I was able to at least get `owl` to show up in the list of packages to be installed before the build process. From this perhaps in can be inferred that `base.v0.15.0` is required by owl rather than `base.v0.14.3`. However, owl is still failing to install because compilation of `base` is failing. A little research led me to discover that this is a [known issue](https://github.com/janestreet/base/issues/126). Perhaps the solution then is to use an older version of `owl` @Sudha247 ?

Sudha247:
Thanks for the PR @moazzammoriani! Looking at the CI, you're spot on about the issue with `base`. Here's a patch to make it work with trunk `opam pin add base.0.15.0 git+https://github.com/patricoferris/base.git#500+random`. I think it's worth trying it, but please note that this is not an official version.

>Perhaps the solution then is to use an older version of owl

I don't think that's going to work, unless the older version of owl doesn't depend on `base`. In any case, it's best to stick with the latest version of `owl` for accurate results.

moazzammoriani:
It appears that this patch doesn't quite suit the needs for this particular problem yet. 

I did gain some insight though into why this patch for `base` didn't work. While testing the patched version of the package on my local environment, opam gave me the following message.

```
[base.0.15.0] synchronised from git+https://github.com/patricoferris/base.git#500+random
base is now pinned to git+https://github.com/patricoferris/base.git#500+random (version 0.15.0)

The following actions will be performed:
  ⊘ remove    owl      1.0.2*             [conflicts with base]
  ↘ downgrade sexplib0 v0.15.0 to v0.14.0 [required by base]
  ⊘ remove    stdio    v0.15.0            [conflicts with base]
  ↘ downgrade base     v0.15.0 to 0.15.0*
===== ↘ 2   ⊘ 2 =====
```
This implies that there are at least two points of failure:

1. `sexplib0` needs to be pinned to `v0.14.0` in the `Makefile` and I'm not sure that is possible since its already pinned to a patched version (https://github.com/ocaml-bench/sandmark/blob/main/Makefile#L136)
2. and `stdio` clearly doesn't play well with this patch either.

Makefile
```diff
@@ -135,9 +135,11 @@ _opam/%: _opam/opam-init/init.sh ocaml-versions/%.json
 	@{ case "$*" in \
 		*5.0*) opam pin add -n --yes --switch $* sexplib0.v0.15.0 https://github.com/shakthimaan/sexplib0.git#multicore; \
 	esac };
-	opam pin add -n --yes --switch $* base.v0.14.3 https://github.com/janestreet/base.git#v0.14.3
+	opam pin add -n --yes --switch $* base.v0.15.0 https://github.com/janestreet/base.git#v0.15.0
+	opam pin add -n --yes --switch $* eventlog-tools https://github.com/ocaml-multicore/eventlog-tools.git#multicore
 	opam pin add -n --yes --switch $* coq-core https://github.com/ejgallego/coq/archive/refs/tags/multicore-2021-09-29.tar.gz
 	opam pin add -n --yes --switch $* coq-stdlib https://github.com/ejgallego/coq/archive/refs/tags/multicore-2021-09-29.tar.gz
+	opam pin add -n --yes --switch $* base.0.15.0 git+https://github.com/patricoferris/base.git#500+random
```
moazzammoriani, line 142: Added the pin to the `Makefile`

.github/workflows/main.yml
```diff
@@ -36,6 +36,16 @@ jobs:
         run: |
           sudo apt-get update && sudo apt-get -y install wget pkg-config libgmp-dev m4 libdw-dev jq python3-pip autoconf jo
           pip3 install intervaltree
+          # Work-around for Ubuntu-OpenBLAS issue for Owl by building OpenBLAS from source
+          sudo apt-get -y install unzip aspcud libshp-dev libplplot-dev gfortran pkg-config git
+          git clone https://github.com/xianyi/OpenBLAS OpenBLAS
+          cd OpenBLAS/
+          make
```
shakthimaan, line 43: Where is `make install` step?

.github/workflows/main.yml
```diff
@@ -36,6 +36,16 @@ jobs:
         run: |
           sudo apt-get update && sudo apt-get -y install wget pkg-config libgmp-dev m4 libdw-dev jq python3-pip autoconf jo
           pip3 install intervaltree
+          # Work-around for Ubuntu-OpenBLAS issue for Owl by building OpenBLAS from source
+          sudo apt-get -y install unzip aspcud libshp-dev libplplot-dev gfortran pkg-config git
+          git clone https://github.com/xianyi/OpenBLAS OpenBLAS
+          cd OpenBLAS/
+          make 
+          cd ..
+          sudo ldconfig /opt/OpenBLAS/lib/
```
shakthimaan, line 45: Why /opt, and why not /usr/local/lib? The other option is to install in /home/opam and update LD_LIBRARY_PATH.

moazzammoriani: I'm actually not sure why /opt is used I just followed the [Dockerfile](https://github.com/owlbarn/owl/blob/main/docker/Dockerfile.ubuntu)

tmcgilchrist:
This needs updating or re-creating from a more recent sandmark version.

PR #371, opened 2022-07-06
New implementation of Graph500 sequential and parallel benchmark (kernel 1 only)

OlivierNicole:
This proposes a new implementation of the first pass (a.k.a. “kernel 1”) of the Graph500 benchmark, namely generating a sparse representation of an undirected graph from a list of edges. It is taken from https://github.com/OlivierNicole/ocaml-graph500. It is adapted from the code of the [graph500seq benchmark](https://github.com/ocaml-bench/sandmark/tree/7cc037472ab67991ef63be6c4e3f6cf83136b120/benchmarks/graph500seq) with a number of optimizations.

Concerns (cc @Sudha247):
- `gen.exe` must be run first as it generates the file `edges.data` which is used by `kernel1Par_run.exe`. The generated file is ~600 MB with the given parameter `-scale 21` (“scale” is the log₂ graph size).
- I haven't tested this new configuration (I'm not sure I can).


OlivierNicole:
For information, on my machine with the given parameters `gen.exe` takes 90 s and kernel 1 takes between 7 s and 0.8 s for 1 and 8 domains, respectively. If we want longer runs we can increase the `-scale` parameter of `gen.exe` (log₂ of the generated graph size, so run time will increase at least exponentially).

Sudha247:
Thank you very much for the PR, @OlivierNicole! 

>gen.exe must be run first as it generates the file edges.data which is used by kernel1Par_run.exe

Maybe we could do something like this - https://github.com/ocaml-bench/sandmark/blob/2e38ae9c160a6aa3eeec69794ed3c439aec7b60e/benchmarks/benchmarksgame/dune, the input files are generated through `dune` rules.

OlivierNicole:
Thanks for the suggestion, I propose something of the sort in the new version.

benchmarks/graph500par/gen.ml
```diff
@@ -0,0 +1,30 @@
...
+    (fun s -> filename := s)
+    "gen.exe [-scale SCALE] [-edgefactor EDGE_FACTOR] OUTPUT_FILENAME";
+  if !filename = "" then begin
+    Printf.eprintf "Must provide graph file argument.\n"; exit 1
+  end;
+  let scale = !scale and edge_factor = !edge_factor in
+  Printf.printf "Generating edge list...\n%!";
+  let pool = T.setup_pool ~num_additional_domains:(!num_domains-1) () in
+  let t0 = Unix.gettimeofday () in
+  let edges = T.run pool (fun () -> Generate.go ~pool ~scale ~edge_factor) in
```
Sudha247, line 26: Since `gen.exe` is generating the graph in parallel, does it make sense to consider this too as a parallel benchmark?

benchmarks/graph500par/generate.ml
```diff
@@ -0,0 +1,102 @@
...
+      if ii_bit then fst_row.(i) <- fst_row.(i) + twopow (ib-1);
+      if jj_bit then snd_row.(i) <- snd_row.(i) + twopow (ib-1);
+    );
+  done;
+  fst_row, snd_row, Array.init m (fun _ -> Random.float 1.)
+
+let permute : 'a array -> 'a array = fun ar ->
+  let with_random_int = Array.map (fun x -> Random.bits (), x) ar in
+  Array.fast_sort (fun (a,_) (b,_) -> compare a b) with_random_int;
+  Array.map (fun (_,x) -> x) with_random_int
```
Sudha247, line 54: I wonder if there's any incentive to making `sort` and `map` parallel. If the `permute` function doesn't take a lot of time, I think it's best to leave it as is.

multicore_parallel_run_config.json
```diff
@@ -111,6 +111,30 @@
...
+      "executable": "benchmarks/graph500par/gen.exe",
+      "name": "graph500_par_gen",
+      "tags": [ "10s_100s" ],
+      "runs": [
+	{ "params": "-scale 21 -edgefactor 16 -ndomains 8 edges.data",
+          "paramwrapper": "taskset --cpu-list 2-13" }
+      ]
+    },
+    {
+      "executable": "benchmarks/graph500par/kernel1Par_run.exe",
```
Sudha247, line 124: Perhaps it's unclear in the README; we need a sequential version of the parallel benchmarks for getting a sequential baseline. Are you able to add a sequential version of this please?

multicore_parallel_run_config.json
```diff
@@ -111,6 +111,30 @@
...
+      "name": "graph500_par_kernel1",
+      "tags": [ "10s_100s", "macro_bench" ],
+      "runs": [
+	{ "params": "-ndomains 2 -nocheck edges.data", "paramwrapper": "taskset --cpu-list 2-13" },
```
Sudha247, line 128: Also an entry with `-ndomains 1`

multicore_parallel_run_config.json
```diff
@@ -111,6 +111,30 @@
       ]
     },
 
+    {
+      "executable": "benchmarks/graph500par/gen.exe",
+      "name": "graph500_par_gen",
+      "tags": [ "10s_100s" ],
+      "runs": [
+	{ "params": "-scale 21 -edgefactor 16 -ndomains 8 edges.data",
```
Sudha247, line 119: If we decide to consider this one as a parallel benchmark, you could add a config similar to `kernel1Par_run.exe`.

Sudha247 (commented):
This is great! Left some few comments below.

benchmarks/graph500par/gen.ml
OlivierNicole: We can certainly consider it to be a parallel benchmark. I'm simply unsure about how to integrate into the machinery the fact that it is a benchmark *and* generates a file used by a later benchmark.

multicore_parallel_run_config.json
OlivierNicole: Done in the updated version.

multicore_parallel_run_config.json
OlivierNicole: Done in the updated version.

multicore_parallel_run_config.json
OlivierNicole: Done in the updated version.

benchmarks/graph500par/gen.ml
OlivierNicole: After checking, I realize that `gen.exe` was already in the list of parallel benchmarks to run. The only remaining uncertainty is side effects and run order, see my comment below.

OlivierNicole:
I believe that I have addressed your comments except the one regarding parallel array mapping and sorting.

Mainly, I added `gen.exe` and `kernel1Seq_run.exe` (and a number of modules they depend on) to `benchmarks/graph500seq`. They are intended as a replacement for `kronecker.exe` and `kernel1.exe`, respectively. However, for now I did not remove those two, which still get run as part of sequential benchmarks for now.

My only uncertainty is about running order and side effects: to run correctly, both `graph500_seq_kernel1` and `graph500_par_kernel1` assume the presence of a file `edges.data` in their working directory, generated by the `gen.exe` benchmark, which I took care to place first in `run_config.json`. This is similar to what is done with `kronecker.exe` and `kernel{1,2,3}.exe`, so I expect it will work.

Sudha247:
Thanks for the updates! CI is green, and the running order makes sense. I'll do a trial run on a local machine to make sure the data files are generated and consumed as intended, and it'd be ready to merge.

benchmarks/graph500par/generate.ml
Sudha247: Maybe something for domainslib. However, we can keep it sequential for now.

moazzammoriani:
I started working on kernel 2 and 3 and used and updated many of the modules implemented in this PR. When I finished working on kernel 2 I created a procedure to print the edgenodes of the graph generated by kernel1 and I ended up getting repeating edges--repo: https://github.com/moazzammoriani/graph500/tree/main/graph500seq. Could this be a bug in `generate.ml`? @OlivierNicole
2022-08-04-194223_1920x651

OlivierNicole:
Hi @moazzammoriani, it's true that I did not try to prevent an edge to be present several times in a node's edge list, because “conceptually” it did not make a difference as to the represented graph, and avoiding duplicates has a slight cost. However I admit for many applications you will want to avoid those duplicates, and I realize now that edge insertion can be done so that the cost is minimal. So feel free to improve my code.

Note that here you have many duplicates because the graph size is the same as the edge factor (16), but in big graphs such duplicates should be rare.

Just to be clear, the fact that an edge is represented at least twice (i.e. at least once in the edge list of its start node, and in the edge list of its end node) is on purpose, so that edge lookup can start from any end.

OlivierNicole:
And to answer your question, IMO this is not a bug in `generate.ml`: as per the Graph500 benchmark rules, the edge list in entry of kernel 1 can be any edge list, possibly with duplicate edges; and the example implementation for edge generation does not prevent duplicates.

shakthimaan:
LGTM!

PR #376, opened 2022-07-19
Refactor nightly scripts to re-use code

punchagan:
- This PR removes extracts duplicated code from nightly_navajo and nightly_turing scripts into a common script.
- Minor refactor of the `run_all_custom.sh` code.

### DIFF
--- a/nightly.sh
+++ b/nightly.sh
@@ -1,0 +1,37 @@
+#!/bin/bash
+
+set -u
+
+# TOKEN required to commit to sandmark-nightly repository
+TOKEN=$1
+
+# The default Sandmark nightly directory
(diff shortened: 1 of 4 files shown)


punchagan:
I removed the code that tries to copy benchmarks from `testing` to `main` branch and the slack notification code.  I'll send that as a separate PR possibly as a GitHub action as discussed with @shakthimaan.  The refactor seems useful, but we possibly need it at this point since 1) the service has been announced publicly and 2) there's only two machines where the nightly runs.

punchagan:
We copy the `nightly_{hostname}.sh` scripts separately to the servers to run the benchmarks. This may not be the best way to configure the server, but with the current setup refactoring common code makes the setup more clunky.

PR #383, opened 2022-07-20
Skip running build if benchmarks exist for a given commit

punchagan:
Closes ocaml-bench/sandmark-nightly-config#6

This check may be too aggressive, i.e., manually running the `run_all_custom.sh` script wouldn't work if a nightly build for that commit already exists, and similarly manual test builds would cause the nightly to skip builds, etc. @shakthimaan let me know if this seems too aggressive to you and if we should skip builds for old commits differently.

### DIFF
--- a/run_all_custom.sh
+++ b/run_all_custom.sh
@@ -47,6 +47,21 @@ find_commit () {
     echo "${COMMIT}" 
 }
 
+is_old_commit () {
+    SEQPAR=$1
+    HOSTNAME=$2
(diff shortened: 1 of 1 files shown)


punchagan:
Updated this PR to check if the "summary.bench" files for a specific commit have been committed on the `main` branch of `sandmark-nightly` to decide if we should skip running builds for it.

shakthimaan:
LGTM!

PR #385, opened 2022-07-21
Make the "soli" bench run longer

xavierleroy:
Currently it takes about 10 milliseconds, which is hard to measure.  This PR adds a command-line parameter to run the test N times.  The suggested values for N in native-code and in bytecode result in execution times on the order of 5 seconds on a fast PC.


Sudha247 (approved):
Looks good to me. Thanks!

PR #386, opened 2022-07-26
Fix incorrect check for whether a commit is old

punchagan:
9dbe92fb09602c628dcfbf278b5426a526fb1fef incorrectly used the output of the
command checking if a bench file existed, instead of the exit code of the
command.  This commit fixes that.


shakthimaan:
LGTM!

PR #389, opened 2022-08-31
Remove macro_bench tag on select benches

### LINKED ISSUE #348: Number of benchmarks are misclassified as macro benchmarks
I recently ran just the macro benchmarks and observed that there are benchmarks that run for 10ms and are classified as macrobenchmarks. The classification that we'd used earlier was that if a benchmark runs for 1s then we'd call it macro.

image

image

We should increase the input size to make the benchmarks run longer. If that is not possible, then the `macro_bench` tag should be removed.

ElectreAAS:
This resolves #348 in the simple way: removing the macro_bench tag from the benches mentionned in the issue.

### DIFF
--- a/run_config.json
+++ b/run_config.json
@@ -113,7 +113,6 @@
       "name": "yojson_ydump",
       "tags": [
         "lt_1s",
-        "macro_bench",
         "run_in_ci"
       ],
       "runs": [
@@ -140,8 +139,7 @@
       "executable": "setrip",
       "name": "setrip",
       "tags": [
-        "1s_10s",
-        "macro_bench"
+        "1s_10s"
       ],
       "runs": [
         {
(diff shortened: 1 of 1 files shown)


shakthimaan:
LGTM!

kayceesrk:
This fix is wrong. As I had mentioned in #348, a number of these have been "fixed" to run longer. This PR removes ones that run longer as well. Did you actually check the latest running time for the mentioned benchmarks and only remove those which ran quicker than 1s (the limit for macro benchmarks)?

Consider a recent sandmark nightly run:

image

The benchmarks that are declassifed as being macro benchmark by this PR are: `yojson_ydump`, `setrip`, `floyd_warshall`, `zarith_pi`, `durand-kerner-aberth`, `soli`, `hamming`. 

Of these, `zarith_pi`, `durand-kerner-aberth`, `soli`, `hamming`, `floyd_warshall` all run longer than 1s in the latest run! Only `setrip` runs slower than 1s. I would revert this PR to only remove `setrip` for now. 

I would recommend reverting this PR and only removing `setrip` (or ideally, make it run longer). 

In general, removing a benchmark from being macro benchmark is an easy task. But adding new ones is extremely hard. So I would be doubly careful if you are relegating something from not being a macro benchmark. Removing benchmarks from the suite adds blind spots to proposed compiler improvements.

PR #343, opened 2022-05-10
Irmin replay benchmark

Sudha247:
Adds the benchmark illustrated in #299. The benchmark uses `irmin.3.2.1`, to build it on `5.0.0+trunk`. This needed pinning of some packages from [alpha repository](https://github.com/kit-ty-kate/opam-alpha-repository).

The benchmark runs for ~20 minutes on an IITM machine.

PR #395, opened 2022-09-14
Irmin replay benchmark

punchagan:
Based on #343 and work by @ElectreAAS and @shakthimaan


Sudha247 (approved):
Looks good to me. Thanks for updating the benchmark!

PR #396, opened 2022-09-15
Detect COMMIT for archives of the form refs/pull

punchagan:
Closes ocaml-bench/sandmark-nightly#90


shakthimaan:
LGTM!

gadmm:
Thanks! Unfortunately, sandmark-nightly still shows e.g. `turing+head+20220919_045431` instead of `turing+HASH+20220919_045431`. Is it just waiting for sandmark-nightly to receive the update?

punchagan:
@shakthimaan It looks like `/home/sandmark/production/run_all_custom.sh` is still the old version and this would need to be updated for this change to take effect. I'm not sure if we have some process in place to update this file or we just update it by hand.

shakthimaan:
Yes, the latest version has now been updated on both Turing and Navajo servers.

PR #400, opened 2022-09-29
Fix execution of Graph500 benchmarks

OlivierNicole:
Currently the benchmarks `graph500par/kernel1_run` and `graph500seq/kernel1_run` fail because they depend on a file `edges.data` which should be generated first. This is an attempt to fix it by generating `edges.data` using dune rules.

### DIFF
--- a/benchmarks/graph500par/dune
+++ b/benchmarks/graph500par/dune
@@ -35,6 +35,11 @@
   (libraries graphTypes generate kernel1Par unix domainslib)
 )
 
+(rule
+  (targets edges.data)
+  (deps (:prog gen.exe))
+  (action (run %{prog} -scale 21 -edgefactor 16 -ndomains 16 %{targets})))
+
 (alias
   (name multibench_parallel)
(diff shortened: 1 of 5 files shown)


shakthimaan:
If you would like to run the benchmarks on the navajo server as well, you can add the entries to https://github.com/ocaml-bench/sandmark/blob/main/multicore_parallel_navajo_run_config.json.

OlivierNicole:
@shakthimaan OK, sure, I added them to that file, too. Thank you for the suggestion.

shakthimaan:
Thanks! Can you also please pull in the latest `bench.Dockerfile` changes for the checks to pass? The latest file is at https://github.com/ocaml-bench/sandmark/blob/main/bench.Dockerfile.

OlivierNicole:
I rebased this PR on the main branch; `bench.Dockerfile` should now be up-to-date.

shakthimaan:
LGTM!

PR #402, opened 2022-10-11
Using sandmark inside current-bench

ElectreAAS:
This PR allows the `make bench` command to run sequential benchmarks and record the results in a way that will be consumed by current-bench.
The last thing needed is to enroll the ocaml-bench organisation into the [github app ocaml-benchmarks](https://github.com/marketplace/ocaml-benchmarksv) if that wasn't done already, and to activate it on the sandmark repository.


shakthimaan:
LGTM!

PR #403, opened 2022-10-12
Update to domainslib.0.5.0

Sudha247:
Some updates to preserve compatibility with `domainslib.0.5.0`.


Sudha247:
Any idea why the CI isn't picking up the latest version of domainslib, even though it's opam file is updated. Does it need to be pinned elsewhere?

shakthimaan:
Do we still need `domainslib.0.4.2`? This version can be removed if domainslib.0.5.0 should be used henceforth.

Sudha247:
#461 subsumes this.

Issue #401, opened 2022-10-07
Add MaPLe benchmarks to Sandmark


shakthimaan:
A number of benchmarks from the MaPLe available at:

  https://github.com/MPLLang/parallel-ml-bench/tree/main/ocaml
  
The benchmarks are also mentioned in the papers by Sam Westrick at:

  https://www.cs.cmu.edu/~swestric/

It will be useful to create an OCaml version of the same for inclusion in Sandmark.

shwestrick:
Would be great to see these in sandmark! The `tokens`, `primes`, `msort_ints`, and `msort_strings` benchmarks are all new. I may add some more benchmarks in the future, too.

The OCaml source at https://github.com/MPLLang/parallel-ml-bench/tree/main/ocaml should be good to go straight out of the box with OCaml 5 and domainslib 0.4.2. The [accompanying library](https://github.com/MPLLang/parallel-ml-bench/tree/main/ocaml/lib) could also be of independent interest...

If anyone has any questions about the benchmarks, let me know. I'm happy to help!

PR #404, opened 2022-10-21
Add (new) mpl benchmarks

Firobe:
Close https://github.com/ocaml-bench/sandmark/issues/401

Add `msort_ints`, `msort_strings`, `primes`, `tokens`, `raytracer` from https://github.com/MPLLang/parallel-ml-bench/tree/main/ocaml/bench to the parallel benchs (with 1, 2, 4, 8, 16, 32 cores each) on Turing and Navajo.


multicore_parallel_navajo_run_config.json
```diff
@@ -18,6 +18,71 @@
...
+      "executable": "benchmarks/mpl/msort_ints.exe",
+      "name": "msort_ints",
+      "tags": ["gt_100s", "mpl"], 
+      "runs": [
+        { "short_name": "1", "params": "-procs 1 -repeat 10 -N 20000000", "paramwrapper": "taskset --cpu-list 5" },
+        { "short_name": "2", "params": "-procs 2 -repeat 10 -N 20000000", "paramwrapper": "taskset --cpu-list 5-6" },
+        { "short_name": "4", "params": "-procs 4 -repeat 10 -N 20000000", "paramwrapper": "taskset --cpu-list 5-8" },
+        { "short_name": "8", "params": "-procs 8 -repeat 10 -N 20000000", "paramwrapper": "taskset --cpu-list 5-11" },
+        { "short_name": "16", "params": "-procs 16 -repeat 10 -N 20000000", "paramwrapper": "taskset --cpu-list 5-19" },
+        { "short_name": "32", "params": "-procs 32 -repeat 10 -N 20000000", "paramwrapper": "taskset --cpu-list 5-35" }
```
Sudha247, line 31: Navajo has 128 cores, so we can spawn 120 domains like the other benchmarks. For example - https://github.com/ocaml-bench/sandmark/blob/main/multicore_parallel_navajo_run_config.json#L188-L217

multicore_parallel_run_config.json
```diff
@@ -19,7 +19,71 @@
...
+      "executable": "benchmarks/mpl/msort_ints.exe",
+      "name": "msort_ints",
+      "tags": ["gt_100s", "mpl"], 
+      "runs": [
+        { "short_name": "1", "params": "-procs 1 -repeat 10 -N 20000000", "paramwrapper": "taskset --cpu-list 5" },
+        { "short_name": "2", "params": "-procs 2 -repeat 10 -N 20000000", "paramwrapper": "taskset --cpu-list 5-6" },
+        { "short_name": "4", "params": "-procs 4 -repeat 10 -N 20000000", "paramwrapper": "taskset --cpu-list 5-8" },
+        { "short_name": "8", "params": "-procs 8 -repeat 10 -N 20000000", "paramwrapper": "taskset --cpu-list 5-11" },
+        { "short_name": "16", "params": "-procs 16 -repeat 10 -N 20000000", "paramwrapper": "taskset --cpu-list 5-19" },
+        { "short_name": "32", "params": "-procs 32 -repeat 10 -N 20000000", "paramwrapper": "taskset --cpu-list 5-35" }
```
Sudha247, line 32: OTOH, Turing has only 28 cores, we can create a maximum of 24 cores.  A sample config for Turing: https://github.com/ocaml-bench/sandmark/blob/main/multicore_parallel_run_config.json#L350-L357. Here are some notes on the machine configurations: https://github.com/ocaml-bench/ocaml_bench_scripts#notes-on-hardware-and-os-settings-for-linux-benchmarking

Sudha247 (changes requested):
Thanks for adding the benchmarks @Firobe!

The config files have the correct benchmarks added to them. But, the `benchmarks/mpl/bench` directory has more benchmarks that are already present in sandmark. Perhaps we can remove the already available ones? You can see the existing parallel benchmarks in `multicore-numerical` and other directories with `multicore-prefix`.

Firobe:
I think it should now be good to go!

Sudha247 (approved):
Thanks for the updates. Looks good to me!

shakthimaan:
Thanks for the review! LGTM.

kayceesrk:
Do these benchmarks appear on the sandmark nightly results? @shakthimaan 

If not, can you make them appear on the nightly results?

Sudha247:
We should add a `macro_bench` tag for them to appear in sandmark-nightly, I think.

kayceesrk:
Do we know that these benchmarks deserve `macro_bech` tag? In particular, do they run for enough time?

kayceesrk:
One way to find out is to add the `macro_bench` tag and then observe the results on the nightly results and tune them appropriately.

Sudha247:
All the benchmarks have a `gt_100s` tag, so I take it as they run for more than 100s at least for the one-core version. In that case, we can safely tag them as `macro_bench`.

kayceesrk:
In addition to the missing `macro_bench` tag, the benchmarks are not in the right format to be visualized. In particular for a benchmark to be visualized, 

1. The benchmark should have a serial version with the name `foo`.
2. The benchmark should have a parallel version with the name `foo_multicore`
3. The first argument (`params` or `short_name`) should be the number of domains. There can be other arguments that follow. When using `short_name`, the format should be `<num_domains>_<other_arg>` with an `_` separating the number of domains and other arguments. 

I also think this is the reason why graph500 results also don't appear in the nightly results.

It appears that this part of the codebase needs a little bit of love and attention. Someone should:

1. Review all of the current parallel benchmarks to see whether each of the benchmarks that appear to be tagged as `macro_bench` does appear in the nightly run. 
2. Make an issue with all the benchmarks tagged as `macro_bench` that doesn't appear in the nightly runs.
3. Triage each one to identify what the problem is.
4. Update the documentation so that the contributors know what is the expectation (see previous list). 
    + Perhaps convert this to a checklist that the contributors and the reviewers can use.
    + Better built tooling that can catch this. 

Unfortunately, the current state is less than ideal. We don't yet have the tools/process to review and catch these before this PR and similar others are merged.

kayceesrk:
> Building a checklist for the contributors and/or reviewers would be a good first step, and then some tooling around making sure at least the easily verifiable checklist requirements are being followed.

Tooling here is a bit tricky. One could possibly build this as a `jq` script that checks for a given benchmark whether serial version exists followed by parallel benchmark named appropriately, etc. 

> graph500 results were disabled because the benchmark implementation was not scalable, and we have disabled the results from showing up in the UI until we fix the scalability.

Thanks for the clarification here.

kayceesrk:
Thanks @Firobe 👍

Firobe:
See https://github.com/ocaml-bench/sandmark/pull/439

Issue #398, opened 2022-09-25
Fix Uuidm.create, Stdlib.Printf.kprintf and ocaml_deprecated_auto_include compiler warnings


shakthimaan:
The follower warnings during sequential benchmarks run build need to be fixed:
```
File "_none_", line 1: 
Alert ocaml_deprecated_auto_include: 
OCaml's lib directory layout changed in 5.0. The unix subdirectory has been
automatically added to the search path, but you should add -I +unix to the
command-line to silence this alert (e.g. by adding unix to the list of
libraries in your dune file, or adding use_unix to your _tags file for
ocamlbuild, or using -package unix for ocamlfind).
File "_none_", line 1: 
Alert ocaml_deprecated_auto_include: 
OCaml's lib directory layout changed in 5.0. The unix subdirectory has been
automatically added to the search path, but you should add -I +unix to the
command-line to silence this alert (e.g. by adding unix to the list of
libraries in your dune file, or adding use_unix to your _tags file for
ocamlbuild, or using -package unix for ocamlfind).
File "benchmarks/lexifi-g2pp/optimization.ml", line 235, characters 39-53:
235 |         | Some f -> (fun evals dist -> Printf.kprintf f "Global Optimizer   Evals: %5i   RMS: %12.8g%%" evals (rms_of_error dist))
                                             ^^^^^^^^^^^^^^
Alert deprecated: Stdlib.Printf.kprintf
Use Printf.ksprintf instead.
File "benchmarks/valet/test_lib.ml", line 19, characters 13-25:
19 |         id = Uuidm.create `V4;
                  ^^^^^^^^^^^^
Alert deprecated: Uuidm.create
Use Uuidm.v instead
File "benchmarks/valet/test_lib.ml", line 46, characters 11-23:
46 |       id = Uuidm.create `V4;
                ^^^^^^^^^^^^
Alert deprecated: Uuidm.create
Use Uuidm.v instead
File "benchmarks/valet/test_lib.ml", line 83, characters 13-25:
83 |     let id = Uuidm.create `V4 in
                  ^^^^^^^^^^^^
Alert deprecated: Uuidm.create
Use Uuidm.v instead
File "benchmarks/valet/test_lib.ml", line 19, characters 13-25:
19 |         id = Uuidm.create `V4;
                  ^^^^^^^^^^^^
Alert deprecated: Uuidm.create
Use Uuidm.v instead
File "benchmarks/valet/test_lib.ml", line 46, characters 11-23:
[...]

PR #411, opened 2022-11-02
Fix warnings on sequential build bench

Firobe:
Close https://github.com/ocaml-bench/sandmark/issues/398

Confirmed there are no warnings left when building sequential benchmarks on Navajo with 5.1.0+trunk


punchagan:
Thanks for fixing these, @Firobe !

shakthimaan:
LGTM!

PR #412, opened 2022-11-07
Remove explicit dependency on a topkg version

punchagan:
topkg is already a transitive dependency and a new release of topkg and updates to it's dependencies caused the build to fail. This commit drops the explicit dependency on topkg and lets opam install it based on the other dependencies.


shakthimaan:
LGTM!

PR #413, opened 2022-11-10
Use dune data_only_dirs stanza instead of sed to ignore benchmarks

punchagan:
Adding and removing directories from this list is easier and much more stable than writing sed rules.  It also moves all the benchmark filtering to a single place and makes it easy to see which benchmarks are being ignored.

We use dune's `data_only_dirs` stanza to tell dune to ignore dune files in some directories. This stanza rule only allows specifying immediate sub-directories. So, we create a `benchmarks/dune` file and add the ignore rules to it.


shakthimaan:
Good enhancement!

PR #414, opened 2022-11-10
Run check_url only in CI instead of every run

punchagan:
If a developer adds new files or changes any thing in the ocaml-versions/*.json files, they can manually run the check_url make target.


shakthimaan:
There are OCaml compiler developers who do clone Sandmark and run locally on their development/test machines, and thus having the check built-in is useful. Adding it to the CI is fine, but, we need to keep it implicit for the `depend` target.

shakthimaan:
LGTM!

PR #415, opened 2022-11-16
Fix typo in mergesort executable name

punchagan:
05f9527357fddf25bc2a5bdebe87bdaed418a124 tried to fix a typo, but introduced another. This commit fixes problems with running mergesort benchmark


shakthimaan:
Thanks for the fix!

PR #418, opened 2022-11-21
Build kernel1_run.exe when running parallel benchmarks too

punchagan:
Fixes an error with running the graph500 benchmarks when running parallel benchmarks.


punchagan:
Merging this to get parallel builds without errors.

PR #419, opened 2022-11-21
Update json formatting to latest of current-bench

ElectreAAS:
As per the (incomplete) code in [the readme of current-bench](https://github.com/ocurrent/current-bench#enroll-your-project), the `metrics` object shouldn't have anything in it that isn't plottable, so no booleans nor nulls nor strings. This PR puts everything of that sorts in the `context` object so it's not lost.


punchagan:
Thanks for working on this, @ElectreAAS! Merged it after minor adjustments to the JSON - removed an extra `,` and the unnecessary `[ ]` around the `metrics` value. Also, added a commit to run `make json` in CI to catch any broken json generation.

PR #417, opened 2022-11-17
Add support for running benchmarks with 4.14 (sequential) and 5.0.0+trunk

punchagan:

PR #421, opened 2022-11-23
Upgrade dune to 3.5.0

punchagan:
This PR removes the special version of dune installation for CI and use the same
installation procedure as actual benchmark runs for CI as well, and upgrades the 
version of dune to 3.5.0

Without this PR, errors occur when trying to install hdr_histogram during the [nightly runs](https://github.com/ocaml-bench/sandmark-nightly/commit/6823fbf8d8fa97a68ee8bf78b5297c1eab402763#diff-b7ac490c32e826964443ee723c3a98f72f454e782fddb1e17c0aa7efeb12f05aR463-R477) (Expand the first diff to jump to the exact lines)

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -40,10 +40,6 @@ jobs:
           sudo apt-get update && sudo apt-get -y install wget pkg-config libgmp-dev m4 libdw-dev jq python3-pip autoconf jo gnuplot-x11 libgmp-dev
           pip3 install intervaltree
           eval $(opam env)
-          opam update
-          opam pin add -n --yes dune 3.5.0
-          opam install dune
-          eval $(opam env)
           export ITER=1
           export OPAM_DISABLE_SANDBOXING=true
           TAG='"run_in_ci"' make run_config_filtered.json
@@ -57,10 +53,6 @@ jobs:
         run: |
           sudo apt-get update && sudo apt-get -y install wget pkg-config libgmp-dev m4 libdw-dev jq python3-pip autoconf jo gnuplot-x11 libgmp-dev
           pip3 install intervaltree
-          eval $(opam env)
-          opam update
-          opam pin add -n --yes dune 3.5.0
-          opam install dune
           eval $(opam env)
           export ITER=1
           export OPAM_DISABLE_SANDBOXING=true
@@ -78,10 +70,6 @@ jobs:
           sudo apt-get update && sudo apt-get -y install wget pkg-config libgmp-dev m4 libdw-dev jq python3-pip autoconf jo gnuplot-x11 libgmp-dev
           pip3 install intervaltree
           eval $(opam env)
-          opam update
-          opam pin add -n --yes dune 3.5.0
-          opam install dune
-          eval $(opam env)
           export ITER=1
           export OPAM_DISABLE_SANDBOXING=true
           TAG='"run_in_ci"' make run_config_filtered.json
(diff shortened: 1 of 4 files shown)


punchagan:
Merging this would make sure we get rid of errors when running the parallel benchmarks, and copy the results to the main branch.

Firobe:
Does this work with OCaml 4.14? In my last PR I tried to not change the overall version of `dune` bacause I remember @shakthimaan telling 4.14 was still tied to `dune` < 3.0

punchagan:
> Does this work with OCaml 4.14? In my last PR I tried to not change the overall version of `dune` bacause I remember @shakthimaan telling 4.14 was still tied to `dune` < 3.0

I'm not sure if it works with 4.14, I haven't tested it yet. But, from the [opam dune page](https://opam.ocaml.org/packages/dune/dune.3.5.0/) it looks like dune itself works with OCaml > 4.08.

Firobe (commented):
If we are upgrading the general version of dune to 3, lines 144-147 in `Makefile` shouldbe removed since we don't need these pins anymore.

Firobe:
I tried on Navajo and 4.14 benchmarks seems broken even without this PR. @shakthimaan can we safely upgrade the `dune` version? Can we ditch OCaml 4.14? Where is 4.14 still used?

If yes, then other than my comment above the PR looks good to me!

punchagan:
Yes, 4.14 is broken. Parallel benchmarks don't run because domainslib doesn't work with it. I have a WIP PR to get serial benchmarks running on 4.14 based on @shakthimaan 's work here: #417.  Since it is just serial benchmarks, I suspect we shouldn't have much trouble with getting things working.  In any case, I think it's valuable to have a build without errors on 5.1.0 and collecting the data from nightly runs to view in the UI.  Any issues with 4.14 can be dealt with separately.

Firobe:
I agree. If needed, another way to deal with the current nightly problem would be to not change anything and install `dune.3.5.0` on the host switch on Turing instead (like done in CI currently), that's what causing the problem to begin with. However I don't see any file describing turing's switch and I can't access it.

punchagan:
> If needed, another way to deal with the current nightly problem would be to not change anything and install `dune.3.5.0` on the host switch on Turing instead (like done in CI currently), that's what causing the problem to begin with. However I don't see any file describing turing's switch and I can't access it.

The problem is not Turing specific and Navajo has the [same problem too](https://github.com/ocaml-bench/sandmark-nightly/commit/805a1c8490d341843719ff57aa1a4ac8f3f4767d#diff-6f236494ed336beb54474bf4cba66f9142814858159074eec7bbc12a3f450d1cR463-R477).   Upgrading to `dune 3.5.0` feels like the right thing to do. Using a different version of `dune` on CI and when running the actual benchmarks is counter-productive and makes it harder for developers to work on Sandmark. 

If we need to make tweaks for OCaml 4.14, we can, since it's already going to have some special stuff to not run parallel benchmarks, etc.

punchagan:
> No! We will need to continue to support 4.14 development branch because 5.0 is not released yet, and we need to give time for projects to migrate.

Sure. We can add back support for 4.14, but currently our nightly builds have errors in the logs and that causes the benchmark runs to be marked as errored runs and are not added on the main branch. I think it would be worth merging these changes and working on 4.14, separately. Thoughts, @shakthimaan ?

punchagan:
We can hold off on merging this. I'm going to rebase the #417 PR to also add support for 4.14 and 5.0.0 with the new dune version, and that PR can be merged.

punchagan:
Closed in favor of #417

PR #423, opened 2022-11-28
Fix bench.Dockerfile to allow running chrt for parallel benchmarks

punchagan:
The Benchmarks section in the README documents this command to correctly run parallel benchmarks. We don't yet run it in Current bench, but this will make us future-proof.


shakthimaan:
LGTM!

PR #426, opened 2022-12-06
Use correct CPU when running benchmarks using current-bench

punchagan:
The CPU is specified using a build arg and currently the configuration is specific to work only for sequential benchmarks. The production worker currently has only 2 isolated CPUs and it is not yet clear how we are going to run the parallel benchmarks using Autumn.


punchagan:
Also, set `BENCH_CPU=2` in `production.conf` on Autumn to get the build working.

shakthimaan:
LGTM!

PR #427, opened 2022-12-06
Remove gnuplot dependency for benchmarks

punchagan:
`gnuplot` was added as a dependecy as a part of the PR #343 since a library for running the Irmin benchmarks depended on it. But, we don't currently run the `irmin` benchmarks and the `gnuplot` dependency installs a whole bunch of X11 dependencies when setting up the OS for running the benchmarks. This makes the setup longer than it needs to be, as of now.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -37,6 +37,7 @@ jobs:
       # Runs a set of commands using the runners shell
       - name: 5.1.0+stable
         run: |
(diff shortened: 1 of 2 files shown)


shakthimaan:
> we don't currently run the irmin benchmarks

My suggestion is to leave a `TODO` comment with the actual dependency name instead of removing it completely. When the ppxlib dependency is resolved, we will run the Irmin benchmarks again.

shakthimaan:
Thanks! LGTM.

PR #432, opened 2022-12-20
Update URLs for OCaml 5.0 and stabilize build by using and older version of runtime_events_tools

punchagan:
The PR updates 5.0 URLs etc after the release last week.

Also, the latest commits to runtime_events_tools add a dependency on `tracing` for supporting the Fuchsia format, but this causes issues with dependency resolution on OCaml 5.1. This PR pins to an older version of runtime_events_tools, while the issue is debugged.


shakthimaan:
There is a type error in the Jupyter notebook in `test_notebooks` for 5.1x:
```
TypeError: ufunc 'isfinite' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''
```

shakthimaan:
LGTM!

PR #433, opened 2022-12-22
Fix regex for getting CONFIG_VARIANT from CONFIG_NAME

punchagan:
The previous regex would parse a name like
`5.1.0+trunk+fabbing+prefetching_like_414` to `5.1.0+trunk\n414`.  This commit makes sure we only look at digits at the begnning of the config name.

*NOTE* The run_all_custom.sh script needs to be copied onto the bench machines.


punchagan:
This would fix broken benchmark runs like [this](https://github.com/ocaml-bench/sandmark-nightly/commit/407550b197f7c5cf552f9828e3c9d7d34fa4551d#diff-057f6c347a8ecb0502f5961ec2dda9d60580e8b788106a8984c258595472a733)

shakthimaan:
Thanks!

PR #434, opened 2023-01-04
Copy RUN_CONFIG_JSON file to results dir

punchagan:
To keep track of the benchmarks that succeeded and failed, it would useful to have the RUN_CONFIG_JSON file with all the benchmark runs configuration. The results bench file can be compared against this to calculate the number of successful and failed benchmarks.

This PR is a complementary PR to https://github.com/ocaml-bench/sandmark-nightly/pull/117


shakthimaan:
LGTM!

PR #437, opened 2023-01-12
Fix wget usage in Makefile

ElectreAAS:
In my search to clean the Makefile, I discovered a bug:
We use `wget` to check for valid urls, but don't exit in case of failure. Adding the exits showcases that in some scenarios, `wget` might not be installed at all, which leads to printing "Error [URL] doesn't exist" even though it does exist, but the system didn't find wget.
The first commit of this PR fixes the printing and the exits, the second fixes the root bug and actually installs wget.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -313,16 +313,22 @@ check_jq:
 	};
 
 check_url: check_jq
-	@{ for f in `find ocaml-versions/*.json`; do					\
+	@{  if [ -z "$$(which wget)" ]; then							\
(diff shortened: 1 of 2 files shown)


ElectreAAS:
Static link to the failing log on current-bench here : http://autumn.ocamllabs.io:8081/job/2023-01-12/155852-ocluster-build-e5e20f
We can see that wget wasn't installed

punchagan:
Thanks for fixing this, @ElectreAAS !

PR #438, opened 2023-01-13
Run the run_all_custom.sh script wrapped in flock

punchagan:
The nightly cron jobs on Navajo have started to take almost a day, and sometimes this could lead to multiple benchmarks getting started at the same time. To prevent this we use the `flock` utility to detect if the process that previously owned a lock file is still alive, and not start another process if that's the case.

The call to `flock` is added to the nightly_* scripts so that it's more visible to the developers, rather than being hidden away in the crontab file.

### DIFF
--- a/nightly_navajo.sh
+++ b/nightly_navajo.sh
@@ -25,7 +25,7 @@ check_sandmark_subdir $SANDMARK_NIGHTLY_DIR
 eval $(opam env)
 
 # Run!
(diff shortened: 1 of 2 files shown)


punchagan:
@shakthimaan can we merge this? Please let me know if you need me to clarify something to help you review and merge this.

punchagan:
> Can you also please add flock to the DEPENDENCIES list in the Makefile?

We can add it here, but we are invoking the `flock` command before we run any make targets. So, not sure it's helpful when flock is not installed on a system.

shakthimaan:
Okay. It is only for the nightly service, so that is fine. LGTM!

PR #439, opened 2023-01-13
Include MPL benchmarks in the nightly pipeline

Firobe:
Fix mpl benchmarks naming conventions and include it in the nightly runs. Additionally, document said conventions.

See https://github.com/ocaml-bench/sandmark/pull/404#issuecomment-1377033072

### DIFF
--- a/README.md
+++ b/README.md
@@ -175,7 +175,8 @@ onto CPU 5.
 The benchmarks also have associated tags which classify the benchmarks. The
 current tags are:
 
-* `macro_bench` - A macro benchmark.
+* `macro_bench` - A macro benchmark. Benchmarks with this tag are automatically
+    run nightly.
 * `run_in_ci` - This benchmark is run in the CI.
 * `lt_1s` - running time is less than 1s on the `turing` machine.
 * `1s_10s` - running time is between 1s and 10s on the `turing` machine.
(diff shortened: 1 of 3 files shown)


kayceesrk:
When this PR is merged, it would be useful to check that the results do appear in the Sandmark Nightly UI. I hope we can remember to do this _after the PR is merged._

shakthimaan:
> it would be useful to check that the results do appear in the Sandmark Nightly UI.

Sure.

> is there a quick way to check (in the CI) whether the additions will run and be visible in the Sandmark nightly UI

These benchmarks are tagged with `macro_bench` and hence they are not run in the CI.

shakthimaan:
LGTM!

punchagan:
The results don't appear in the UI, since the serial version of the benchmarks is missing. 

@Firobe as mentioned by KC [here](https://github.com/ocaml-bench/sandmark/pull/404#issuecomment-1377033072)

```quote
The benchmark should have a serial version with the name `foo`.
The benchmark should have a parallel version with the name `foo_multicore`
```

So, the config files need to have serial runs for each of the MPL benchmarks, `msort_ints`, `msort_strings`, etc.  (Again, as mentioned by KC in the comment referenced above, *Parallel version running on 1 domain is not the same as the sequential version*)

PR #441, opened 2023-02-23
Remove multicore-lockfree benchmarks

punchagan:
The benchmarks are not being run on Sandmark nightly runs, and are out-dated. It seems like a better idea to run them using current-bench. Removing them completely instead of skipping them reduces the maintenance burden on the sandmark maintainers. They can always be revived from version control, if required.


Sudha247 (approved):
Looks good, thanks!

PR #444, opened 2023-04-14
Better Git defaults

MisterDA:
Ignore `.DS_Store` files, specific to macOS.
Force LF line endings on shell scripts.


punchagan:
Thanks @MisterDA for taking the time to fix this! :tada:

PR #445, opened 2023-04-18
Fix OCaml trunk builds

punchagan:
OCaml trunk version has been bumped up to 5.2 and this PR fixes broken sandmark builds.


shakthimaan:
LGTM!

Issue #446, opened 2023-04-24
A number of parallel benchmarks seem to be broken


kayceesrk:
![image](https://user-images.githubusercontent.com/410484/233893735-310e4d1f-8e37-49b9-bd0d-98079b7e52e5.png)

[Corresponding sandmark nightly run](https://sandmark.ocamllabs.io/?app=Parallel+Benchmarks&parallel_00=5.2.0%2Btrunk%2Bparallel&parallel_10=5.2.0%2Btrunk%2Bfix-idle-gc%2Bparallel&parallel_12=turing_5.2.0%2Btrunk%2Bfix-idle-gc%2Bparallel_20230423_7440f24&parallel_02=turing_5.2.0%2Btrunk%2Bparallel_20230423_063894d&parallel_find_by=variant&parallel_num_variants=2&parallel_01=20230423&parallel_11=20230423)

punchagan:
The benchmarks that are broken here are the MPL benchmarks added in #404 & #439. The benchmarks are broken due to missing serial versions of the benchmarks as noted in the comment here: https://github.com/ocaml-bench/sandmark/pull/439#issuecomment-1400252463

shwestrick:
It should be easy to generate a sequential version of these benchmarks. Just need to replace the [`ForkJoin`](https://github.com/ocaml-bench/sandmark/blob/main/benchmarks/mpl/lib/Forkjoin.ml) module with an equivalent sequential implementation. All parallelism goes through that module.

E.g. `let par f g = (f(), g())` and similarly for the other functions.

PR #449, opened 2023-05-03
Basic tooling to validate multicore_*_run_config.json

punchagan:
This PR adds some basic tooling to validate benchmarks configurations added to `multicore_*_run_config.json` files.  We also add a CONTRIBUTING.md file that gives some basic instructions on adding a parallel benchmark. 

The PR also modifies the existing JSON files to fix validation errors. This validation removes the benchmarks listed in #446 from the nightly runs. The run_config.json files need to be triaged separately to see why some of the benchmarks are skipped from nightly runs (don't have the `macro_bench` tag), and need to be fixed.

### DIFF
--- a/.github/workflows/validate-multicore-config.yml
+++ b/.github/workflows/validate-multicore-config.yml
@@ -1,0 +1,22 @@
+---
+name: Validate Multicore Config
+
+on:
+  pull_request:
+    branches: [main]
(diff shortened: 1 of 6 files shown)


kayceesrk (commented):
Looks great. Thanks.

PR #452, opened 2023-05-29
Print the JSON for each pausetimes benchmark in a single line

punchagan:
For other kinds of benchmark measurements, the IPython notebooks and the nightly UI Python code assumes that each line in the benchmark file is valid JSON.  For consistency, it would help the pausetimes data formatted similarly.


ElectreAAS (approved):
LGTM!

PR #457, opened 2023-06-05
Add @stedolan `markbench` benchmark for prefetching

fabbing:
This PR, a joint effort with @MisterDA, adds a very slightly modified version of @stedolan `markbench` micro-benchmark.
This micro-benchmark was first used in https://github.com/ocaml/ocaml/pull/10195 and then in https://github.com/ocaml/ocaml/pull/11827 to validate prefetching speedup while the GC is tracing blocks.

It could be useful as a sort of regression test running in Sandmark.

It would be preferable to use the reported seconds/GC time as calculated by the benchmark, which is reported on `stdout`, since it avoids accounting for setup time.
How could this could be achieved in Sandmark?

### DIFF
--- a/benchmarks/markbench/dune
+++ b/benchmarks/markbench/dune
@@ -1,0 +1,6 @@
+(executable
+ (name markbench)
+ (libraries unix)
+ (modes native))
+
+(alias (name buildbench) (deps markbench.exe))
--- a/benchmarks/markbench/markbench.ml
+++ b/benchmarks/markbench/markbench.ml
@@ -1,0 +1,39 @@
+(* Micro benchmark by Stephen Dolan to monitor prefetching performance for
+   regressions *)
+
+type t = {
+  mutable a : int option;
+  mutable b : float;
+  mutable c : float;
+  mutable d : float;
+  mutable e : float;
+  mutable f : float;
+}
+
+let make_arr () =
+  let nums = Array.init (1 lsl 20) (fun i -> Some i) in
+  let a = Array.init 1_024 (fun _ -> []) in
+  for i = 1 to 10_000_000 do
+    let n = ((i * 34872841) lsr 13) land 0x3ff in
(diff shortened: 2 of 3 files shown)


punchagan:
Thanks for the contribution, @fabbing and @MisterDA. As discussed in person, the benchmarks are currently run using different wrappers, like `orun`, `perfstat` or `pausetimes`. 

But, the use case of being able to measure specific parts of a program as part of a benchmark makes sense to me. We do have such benchmarks being run on different repositories, using `current-bench`, which allows repositories to define their own custom benchmarks. I wonder if maybe that is a better place to have this benchmark? Or if we should add support for doing these kinds of benchmarking with micro-benchmarks in Sandmark. 

@shakthimaan or @kayceesrk might have thoughts on this.

kayceesrk:
> But, the use case of being able to measure specific parts of a program as part of a benchmark makes sense to me.

Sandmark is not built for measuring and reporting on specific parts of the program. All the reported metrics are for the entire program. Breaking this invariant complicates how we measure, report and analyse the benchmarks in Sandmark. I am not keen on breaking this.

`current-bench` is probably the better place. The other option is to make the prefetching specific bits run much longer so that its effects dominate the program behaviour measured as a whole. This may be as simple as running the core parts of the algorithm repeatedly so that prefetching effects are magnified.

fabbing:
> `current-bench` is probably the better place. The other option is to make the prefetching specific bits run much longer so that its effects dominate the program behaviour measured as a whole. This may be as simple as running the core parts of the algorithm repeatedly so that prefetching effects are magnified.

The core part is actually already run several times (and can be easily tweaked with the launch parameter).
The GC/tracing time of the benchmark largely dominates the setup time, so the benchmark can be kept as it is in Sandmark.

kayceesrk:
Sounds great.

PR #464, opened 2024-01-16
Add Owl GC regression test case.

tmcgilchrist:
Originally taken from
https://gitlab.inria.fr/fangelet/ocaml-gc-regression-examples and supplied by Florian Angeletti (@Octachron). This code was supplied on [Performance regression with multiple domains between 5.0 and 5.1](https://github.com/ocaml-bench/sandmark/compare/main...tmcgilchrist:owl_gc_regression?expand=1#top). 

This test will output timings to stdout of the form `4.14.1 | 2.395s |  1949 |   487 ` representing   `compiler version, time, minor GC, major GC`. How will that output be captured by sandmark or is there instrumentation already to capture this sort of GC information?

### DIFF
--- a/run_config.json
+++ b/run_config.json
@@ -1546,6 +1546,19 @@
           "params": "20"
         }
       ]
+    },
+    {
+      "executable": "benchmarks/owl/owl_gc.exe",
+      "name": "owl_gc",
+      "tags": [
+        "1s_10s",
+        "macro_bench"
+      ],
+      "runs": [
+        {
+          "params": ""
+        }
+      ]
     }
   ]
 }
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -36,7 +36,7 @@ jobs:
 
       - name: Install dependencies
         run: |
-          sudo apt-get update && sudo apt-get -y install wget pkg-config libgmp-dev m4 libdw-dev jq python3-pip autoconf jo libgmp-dev
+          sudo apt-get update && sudo apt-get -y install wget pkg-config libgmp-dev m4 libdw-dev jq python3-pip autoconf jo libgmp-dev libopenblas-dev liblapacke-dev zlib1g-dev
           pip3 install intervaltree
 
       # Runs a set of commands using the runners shell
@@ -93,7 +93,7 @@ jobs:
 
       - name: Install dependencies
         run: |
-          sudo apt-get update && sudo apt-get -y install wget pkg-config libgmp-dev m4 libdw-dev jq python3-pip autoconf jo libgmp-dev
+          sudo apt-get update && sudo apt-get -y install wget pkg-config libgmp-dev m4 libdw-dev jq python3-pip autoconf jo libgmp-dev libopenblas-dev liblapacke-dev zlib1g-dev
           pip3 install intervaltree
 
       # Runs a set of commands using the runners shell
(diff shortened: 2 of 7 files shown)


punchagan:
Thanks for taking the time to contribute this PR!

I've updated this PR to correctly install the dependencies for owl.  The problem was that the `opam` file changes had `owl` and `owl-base` version specified incorrectly with the `=` missing ( `"owl" {"1.1"}` instead of  `"owl" {= "1.1"}` ). I've also added missing OS-level dependencies.  So, the benchmark dependencies are installed, and the code compiles.

But, I've been unable to run the benchmark itself (on `navajo`).  

```
punchagan@navajo:~/sandmark/_build/5.2.0+trunk_1/benchmarks/owl$ ./owl_gc.exe 
Fatal error: exception Invalid_argument("index out of bounds")
```

> This test will output timings to stdout of the form `4.14.1 | 2.395s | 1949 | 487 ` representing `compiler version, time, minor GC, major GC`. How will that output be captured by sandmark or is there instrumentation already to capture this sort of GC information?

We use [`orun`](https://opam.ocaml.org/packages/orun/) as a wrapper to capture this information. So, you wouldn't need to do anything additionally in the benchmark code.

Octachron:
Rather than adding empty parameters, I think it is better to remove the reporting part from the benchmark (the last `let _ = ...`) since sandmark will take care better of this aspect.

punchagan:
> Rather than adding empty parameters, I think it is better to remove the reporting part from the benchmark (the last `let _ = ...`) since sandmark will take care better of this aspect.

The reporting part from the benchmark can be removed, yes. :+1: 

The last commit adds empty params field to the  `run_config.json`, since the sandmark [rungen](https://github.com/shakthimaan/rungen) library expects there to be a `params` field. I'm not sure if you are referring to that change or something else.

Octachron:
Sorry, I misread the last commit as a fix for the invalid argument error which comes from the report section of the benchmark.

punchagan:
> Sorry, I misread the last commit as a fix for the invalid argument error which comes from the report section of the benchmark.

I've squashed the last commit into the first one; And also removed the reporting part of the benchmark. I think we should be good to merge this, once the github action finishes.

run_config.json
```diff
@@ -1546,6 +1546,19 @@
...
         }
       ]
+    },
+    {
+      "executable": "benchmarks/owl/owl_gc.exe",
+      "name": "owl_gc",
+      "tags": [
+        "10s_100s",
+        "macro_bench"
+      ],
```
tmcgilchrist, line 1556: @punchagan I wasn't sure what tags make sense here. I copied them from `benchmarks/markbench/markbench.exe` in the section above. Can you confirm they are correct?

punchagan: Based on a test run on `turing`, I changed the `10s_100s` tag to `1s_10s` tag. I see that it can be tricky to add these timing tags, when adding a new benchmark. I don't know how they've been historically added. 

```json
{
  "name": "owl_gc.",
  "command": "taskset --cpu-list 5 ./owl_gc.exe",
  "time_secs": 4.113397836685181,
  "user_time_secs": 3.971614,
  "sys_time_secs": 0.119988,
  "maxrss_kB": 23844,
  "ocaml": {
    "version": "5.2.0+dev2-2023-12-22",
    "c_compiler": "gcc",
    "native_ldflags": "",
    "native_compiler": "true",
    "architecture": "amd64",
    "word_size": "64",
    "system": "linux",
    "function_sections": "true",
    "tsan": "false",
    "supports_shared_libraries": "true",
    "native_dynlink": "true",
    "naked_pointers": "false",
    "linear_magic_number": "Caml1999L033"
  },
  "gc": {
    "allocated_words": 15009670,
    "minor_words": 14802875,
    "promoted_words": 92633,
    "major_words": 299428,
    "minor_collections": 10921,
    "major_collections": 10921,
    "forced_major_collections": 0,
    "heap_words": 102167,
    "top_heap_words": 102570,
    "mean_space_overhead": 694.967482
  },
  "codesize": 2461953,
  "ocaml_url": "https://github.com/ocaml/ocaml/archive/refs/heads/5.2.zip"
}
```

tmcgilchrist:
Thanks @punchagan

PR #466, opened 2024-01-30
Remove pins for sexplib0 and base libraries

punchagan:
They work with the latest versions of OCaml, and needn't be pinned to older versions or special branches/forks


shakthimaan:
LGTM!

Issue #471, opened 2024-06-04
Benchmark runs seem to be failing due to changes on trunk


punchagan:
While working on a fix for #470, it looks like another breaking change has been made in https://github.com/ocaml/ocaml/commit/51e5cf226b39986abf03cbaad47cf49a1c35bc69. Packages like `base64.3.4.0` fail to compile/install. 

```sh
#=== ERROR while compiling base64.3.4.0 =======================================#
# context     2.2.0~beta2 | linux/x86_64 | ocaml-variants.5.3.0+trunk | https://opam.ocaml.org/#b502ec76
# path        ~/.opam/ocaml-ocaml-51e5cf226b39986abf03cbaad47cf49a1c35bc69/.opam-switch/build/base64.3.4.0
# command     ~/.opam/opam-init/hooks/sandbox.sh build dune build -p base64 -j 31
# exit-code   1
# env-file    ~/.opam/log/base64-436405-1a5bf6.env
# output-file ~/.opam/log/base64-436405-1a5bf6.out
### output ###
# File "src/dune", line 7, characters 0-126:
#  7 | (rule
#  8 |  (targets unsafe.ml)
#  9 |  (deps (:config ../config/config.exe) unsafe_pre407.ml unsafe_stable.ml)
# 10 |  (action (run %{config})))
# (cd _build/default/src && ../config/config.exe)
# Error: variable "ocamlc_cflags" not found in the output of `/home/punchagan/.opam/ocaml-ocaml-51e5cf226b39986abf03cbaad47cf49a1c35bc69/bin/ocamlc.opt -config`
```

PR #472, opened 2024-06-04
Pin dune and base to work around ocaml trunk changes 

punchagan:
The primary changes in this PR are:

- Pin `base` to workaround Format_doc change (see commit message)
- Pin `dune` to workaround splitting native and bytecode compiler flags (see commit message)

The PR also has other small changes:

- upgrades `dune` to latest version
- Switches OCaml 5.2.0 to 5.2.1
- Fixes broken notebook tests by pinning Python dependencies
- Disable sys dune hack for nightly builds


punchagan:
CI for OCaml 5.3 is expected to fail due to #471

punchagan:
Pushed a couple of new commits to use a patched version of `dune` to work around the changes in https://github.com/ocaml/ocaml/pull/13201 where the compiler flags for `bytecode` and `native` compilers have been separated out, and name prefixes have been changed to `bytecode_` and `native_` from `ocamlc_` and `ocamlopt_`.

punchagan:
@shakthimaan is my understanding correct that dune would previously not compile against `multicore` branch and `USE_SYS_DUNE_HACK` was required? It seems okay to me to disable this hack and use a patched version of `dune`. Can you please confirm with a review? Thanks!

shakthimaan:
Sure. It will be good to decide on LTS and whether you need to support earlier versions. Otherwise, it is better to upgrade.

punchagan:
Deployed `run_all_custom.sh` script to `navajo`. Remote login to `turing` fails - deployment pending.

punchagan:
[Builds](https://github.com/ocaml-bench/sandmark-nightly/commit/0b597482a68d213688d16c17b0c70156c29c0613) on navajo are running again!

PR #480, opened 2025-11-04
ahrefs-devkit utils GC benchmarks

koonwen:


tmcgilchrist:
Thanks for the contribution @koonwen I'm busy with other things right now but I should get time to look at this in 2 weeks. Could you provide some more context on what these benchmarks are? Are they representative of particular workloads at ahrefs?

koonwen:
Hey @tmcgilchrist, these are some benchmarks covering some of the more memory intensive utils of our opensource [devkit](https://github.com/ahrefs/devkit) library. The benchmarks are pretty generic and not specific to any workload but presents a large target that will impact us if there are any GC improvements/regressions since most of the code we build depends on this library. (i.e. the html parser and compression/decompression utils are heavily used by our crawler).

We care mostly about the GC benchmarks between 4.14 and 5.4 (and above) as we are working on this migration. So we're not concerned about performance earlier than 4.14

*disclaimer, the benchmarks were AI generated but I had a look through and they look reasonable

avsm:
This is great to see, @koonwen! Just a thank you for submitting it...

tmcgilchrist:
Thanks for the info, please get in touch on slack or email if you need specific help with that migration.

koonwen:
@tmcgilchrist, I've had a second look to simplify the code and the deps. Also tested this against ocaml.4.14.0 and 5.4.0 and they build correctly. Let me know if anything else needs to be done

tmcgilchrist:
This would be another interesting example to add to the benchmarking suite  @udesou similar to the markbench example.

